PR #3, opened 2023-09-28
Implementing some elementary data quality tests

da115115:
* References:
  + Issue on the `datacontract.com` CLI utility: https://github.com/datacontract/cli/issues/2
  + Datacontract.com section on https://github.com/data-engineering-helpers/data-contracts/tree/main/datacontract.com
* [Simon](https://github.com/saugerDecathlon) and I are working together
* My goal, for now, is to implement a very first draft of checking the data quality with Soda Core, like documented in https://github.com/data-engineering-helpers/data-contracts/tree/main/datacontract.com#check-the-data-quality-of-a-table-thanks-to-a-data-contract


stefannegele:
Nice to see you getting started! Just write, if I can help with anything!

da115115:
> Nice to see you getting started! Just write, if I can help with anything!

Thanks!

I would like to implement a simple use case/example:
* Reference: https://github.com/data-engineering-helpers/data-contracts/tree/main/datacontract.com
  + Data contract: https://github.com/data-engineering-helpers/data-contracts/blob/main/datacontract.com/contracts/data-contract-flight-route.yaml
  + Quality specification part (cross-referenced in the data contract): https://github.com/data-engineering-helpers/data-contracts/blob/main/datacontract.com/contracts/data-contract-flight-route-quality.yaml
  + The quality type is `SodaCL`
  + The quality checks pass with the following command: `soda scan -d duckdb_local -c soda-conf.yml contracts/data-contract-flight-route-quality.yaml` (once the `db.duckdb` DuckDB database file has created with `duckdb db.duckdb < sql/duckdb-ddl-create-view-from-csv.sql`
* The `datacontract` CLI should redirect the quality checks to Soda Core when the type of the quality specification is `SodaCL` . I just had a look on how to call Python functions from Go code (as Soda Core, like Great Expectations, is coded in Python), and that does not appear obvious (https://github.com/go-python/cpy3 seems to support only Python 3.7 and had no activity for 1.5+ year). So, the Go code could call the Soda Core CLI with a simple Shell/system call.

There is a simple thing where you could maybe help, @stefannegele : writing the part (in https://github.com/data-engineering-helpers/datacontract-cli/blob/main/quality.go) where the specification is cross-referenced from another YAML file; which is the case in my example: https://github.com/data-engineering-helpers/data-contracts/blob/main/datacontract.com/contracts/data-contract-flight-route.yaml#L59
Right now, https://github.com/data-engineering-helpers/datacontract-cli/blob/main/quality.go parses only the data contract YAML file itself, not the cross-referenced YAML file. I do not know whether you already implemented that part. For https://github.com/data-engineering-helpers/datacontract-cli/blob/main/quality.go , I took inspiration from https://github.com/data-engineering-helpers/datacontract-cli/blob/main/schema.go and https://github.com/data-engineering-helpers/datacontract-cli/blob/main/lint.go

da115115:
If you need contributor access to the https://github.com/data-engineering-helpers GitHub organization, do not hesitate. Just let me know and I'll add you to one of the teams (https://github.com/orgs/data-engineering-helpers/teams).

stefannegele:
Hi @da115115 , thanks for the insights. I also think integrating SODA via a command line execution is fine for now. We will figure out a good solution, while integrating other tools.

I will put the resolution of referenced resources in my to-do list!

stefannegele:
@da115115 
with https://github.com/datacontract/cli/commit/d14490d6c90f8ca208266e1c3cc59962ca7d99ff remote and local references are resolved when using `GetValue`, e.g. Is that what you needed?

da115115:
> @da115115 with [d14490d](https://github.com/datacontract/cli/commit/d14490d6c90f8ca208266e1c3cc59962ca7d99ff) remote and local references are resolved when using `GetValue`, e.g. Is that what you needed?

From what I can see in the Go code, yes, that is exactly what I meant. I have yet to test it, especially the part where a remote reference is downloaded locally. As a matter of fact, the same way there is a default file-name for local data contract files (that is `datacontract.yaml`), we could have default fine-names for cross-referenced schema files (eg, `datacontract-schema.yaml`) et quality files (eg, `datacontract-quality.yaml`). That way, when a remote data contract (specified by a URL) itself cross-references schema and/or quality files, those could be downloaded locally into those local files.

stefannegele:
We have just released version 0.3.0. The new `inline` command might also be helpful for you ;) 

Tell me, if I can help to merge my latest changes.

da115115:
> We have just released version 0.3.0. The new `inline` command might also be helpful for you ;)
> 
> Tell me, if I can help to merge my latest changes.

Thanks!

I've just tried the new `inline` method, but it did not work that much.

Here's what I've tried:
1. Setup
```
mkdir -p ~/dev/infra/data-contracts && \
  git clone https://github.com/data-engineering-helpers/data-contracts.git ~/dev/infra/data-contracts/data-contracts && \
  cd ~/dev/infra/data-contracts/data-contracts/datacontract.com
```
2. `inline` command, not working as expected (from my point of view; hopefully I'm wrong)
```
  datacontract quality --file contracts/data-contract-flight-route.yaml
  cat contracts/data-contract-flight-route.yaml # shows that the $ref are still there, nothing has been inlined
```
2.2. Revert local changes
```
git checkout -- .
```
3. Quality checks
```
datacontract quality --file contracts/data-contract-flight-route.yaml 
$ref contracts/data-contract-flight-route-quality.yaml
```

da115115:
And on this branch/pull request, I've merged the last changes, up until 0.3.0, and there are now a few compilation issues:
```
go build -o datacontract
# github.com/datacontract/cli
./quality.go:9:27: undefined: ReadLocalDataContract
./quality.go:10:23: undefined: ParseDataContract
./quality.go:30:1: missing return
```
I guess the `ReadLocalDataContract()` and `ParseDataContract()` functions have been renamed. I can of course investigate on my own; but since you made the change, you may see what's wrong much quicker than what I'll be able to for now.

Thanks!

stefannegele:
> I've just tried the new inline method, but it did not work that much.
I tried the command on [this](https://github.com/data-engineering-helpers/data-contracts/blob/main/datacontract.com/contracts/data-contract-flight-route.yaml) file. Your "$ref ..." should be "$ref: ..." (adding colons). Then it works.

> I guess the ReadLocalDataContract() and ParseDataContract() functions have been renamed. 
Correct. You now can just use `GetDataContract()`. This will give you a `DataContract` object, therefor usage of a specific `ParseDataContract()` is not needed any longer.

da115115:
> > I've just tried the new inline method, but it did not work that much.
> 
> I tried the command on [this](https://github.com/data-engineering-helpers/data-contracts/blob/main/datacontract.com/contracts/data-contract-flight-route.yaml) file. Your "$ref ..." should be "$ref: ..." (adding colons). Then it works.

Thanks, indeed the inlining now seems to work better. However, if I follow up with a `lint` command, there is an internal error:
```
datacontract inline --file contracts/data-contract-flight-route.yaml 
datacontract lint --file contracts/data-contract-flight-route.yaml 
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x0 pc=0x137e157]
```

Moreover, it may be better, when executing the `inline` command, to produce another file, rather than overwriting the current data contract (specified by the `--file` parameter). We could maybe have more specific parameters like `--from-file` and/or `--to-file`, à-la-Unix, where every command can take as input a file (which may be `stdin`) and produce as output a file (which may be `stdout`).

stefannegele:
I just added some tickets. Does not mean I will implement it right now, but it's open for discussion.

The bug with lint, I will check when I have time - probably end of the week. I wanted to do some cleanup and add better testing anyways.

stefannegele:
@da115115 can you please give more info on https://github.com/datacontract/cli/issues/5 (add it to the ticket, please)? I can't replicate it.

file.go
```diff
@@ -0,0 +1,36 @@
...
+
+import (
+	"fmt"
+	"os"
+	"path/filepath"
+	"log"
+	"code.cloudfoundry.org/bytefmt"
+)
+
+func BrowseDir(dirname string) error {
```
stefannegele, line 11: Can you explain the concrete use case for this? It adds a dependency, so we need to be sure to really need it. If we need it, avoiding the dependency would be great.

main.go
```diff
@@ -8,6 +8,8 @@ import (
 )
 
 const dataContractFileName = "datacontract.yaml"
+const qualitySpecFileName = "datacontract-quality.yaml"
```
stefannegele, line 11: Shouldn't we stick with the Syntax we already introduced? External files via "$ref:" and also allowing to embed yaml/strings? We cannot expect that all quality definitions are in yaml format. Can I maybe help with that?

main.go
```diff
@@ -101,13 +117,6 @@ func main() {
 
 					return Lint(ctx.String("file"), ctx.String("schema"))
 				},
-			}, {
-				Name:  "test",
```
stefannegele, line 105: I hope it's okay if we stay with the command names we decided on internally. We wanted to adopt the dbt syntax and name this command `test`. I know, this might unintuitive for you as Soda users.

One reason behind this, is to keep the commands small.

main.go
```diff
@@ -119,18 +128,65 @@ func main() {
...
+						Value: "quality.specification",
+						Usage: "definition of a custom path to the quality specification in your data contract",
+					}},
+				Action: func(ctx *cli.Context) error {
+					pathToSpecification := strings.Split(ctx.String("quality-specification-path"), ".")
+					return PrintQuality(ctx.String("file"), ctx.String("quality-file"), pathToSpecification)
+				},
+			},
+			{
+				Name:  "quality-init",
```
stefannegele, line 147: Can this be integrated into `test`/`quality-check`?

You can probably prepare the quality folder and copy the checks inside as a temporary file, which you can delete after the execution.

stefannegele (commented):
Hi @da115115, nice work!
I had a brief look at your changes and added some comments. Please tell me if I can help or if I misunderstood something.

main.go
da115115: Ok, that is fine. Fixed in [6c0218e](https://github.com/datacontract/cli/pull/3/commits/6c0218eea40c942872fa9fdfc09a8415a1f32d8c)

As you say, I do not find it so intuitive to have a `quality` command just printing the quality check specification, and a `test` command to check the quality, whereas in many software utilities, the test command is usually reserved for testing purpose of the software itself. All is a question of proper documentation, so it is fine :)

main.go
da115115: If you know how to do that, do not hesitate.
The point at stake is to allow several/many quality checks, without having to specify/prepare the quality checks (_e.g._, with Soda configuration) every time: we prepare the quality checks once (as it may take a long time) and we check the quality of data many times.

Note that, in the current version of `sodaQuality.go`, the preparation/initialization step just creates the `quality` temporary directory, if needed, and writes a YAML Soda configuration file so as to use a local DuckDB database to store the data sets to be checked.
The users have then to create the local DuckDB database, which stores the data sets to be checked against the rules specified in the data contract. That local DuckDB database file may be copied from some other places. Such a procedure is documented in https://github.com/data-engineering-helpers/data-contracts/tree/main/datacontract.com#check-the-data-quality-of-a-table-thanks-to-a-data-contract
That step may be time-consuming. That is why I preferred not to include the initialization step in the `test` command.
On the other hand, since the initialization step is currently not destroying any previous work, if already existing, it would NOT be harmful (destructive of previous work, wasteful of time/resources) to systematically include the initialization step in the `test` command.
I let you decide what you prefer. Just let me know and I can include the initialization step in the Soda quality check (`test`) command.

file.go
stefannegele: Thanks!

main.go
da115115: That one would need a dedicated issue, I guess.

Locally, we may prefer to have distinct specification files (whether YAML or any other format, you are right), each dedicated to the respective parts of the contracts (_e.g._, schema, quality checks, input, output). Those dedicated specification files are natural when the remote data contract cross-references those parts; the `init` command could then pull/download locally all the specification parts in local dedicated specification files, whose file-names would correspond to the remote cross-references.

The question is: what should we do locally when the specification part is inline? A solution is to nevertheless create a local dedicated file replicating that specification part, and leave the data contract untouched.
That way, when using/wrapping external tools (such as Soda or Great Expectations), those latter can use directly the dedicated specification files: the whole keeps simple, straightforward.
We can also replace in the local data contract the specification part by a cross-reference on the local dedicated specification file. It avoids duplication of the specification, but it seems more fragile to me (while being cleaner).

The whole point is to have locally dedicated specification files. As long as we have that, I am fine with the solution :)

main.go
stefannegele: > Note that, in the current version of [sodaQuality.go](https://github.com/data-engineering-helpers/datacontract-cli/blob/main/sodaQuality.go), the preparation/initialization step just creates the quality/ temporary directory, if needed, and writes a YAML Soda configuration file so as to use a local DuckDB database to store the data sets to be checked.

That's what I saw, and why I'm asking. I'm actually not sure, if the CLI should be responsible for setting up the test environment. Since Soda supports several other data sources next to DuckDB, we would need to aim to support all of them. And probably the same for other tools. In my opinion, the datacontract CLI should expect a completely set up data quality tool and just run the configured checks on it - in this case, copy the checks and run them in Soda as you already implemented it. Please correct me, if you think I'm on the wrong path here.

For the sake of simplicity, I think it's best to integrate everything into `test` for now.

> If you know how to do that, do not hesitate.

I'm not sure what this applies to? [The other comment](https://github.com/datacontract/cli/pull/3#discussion_r1357876873)?

main.go
stefannegele: The thing is, we want our data contract to know about all quality specifications. Ideally, they are even embedded in the file, so everyone looking at the contract, can see what qualities apply to the data. If you allow injecting the quality file from outside, you make the data contract knowing about it obsolete.

stefannegele: Would it help to allow multiple quality specifications in the data contract specification?

main.go
da115115: [17f321f](https://github.com/datacontract/cli/pull/3/commits/17f321fa7249949b278b054123b3b8011ed9225e)
The initialization step has been embedded into the quality check part, and that holds true only for the Soda wrapper for now. As a matter of fact, other quality checkers may not need initialization steps.

da115115: > I'm not sure what this applies to? https://github.com/datacontract/cli/pull/3#discussion_r1357876873?

I was referring to the following comment:

> Can this be integrated into `test / quality-check`?

Indeed, I thought you were meaning, with the slash (`/`) character, some sort of levels in the commands, as the slash (`/`) is usually the separator in file-system folder structures.
I now realize that the (`/`) character was here meaning "or" and/or _aka_ (also known as): `test` or/aka `quality-check`. So, it is all clear by now, thanks!

da115115: > [...] I'm actually not sure, if the CLI should be responsible for setting up the test environment. Since Soda supports several other data sources next to DuckDB, we would need to aim to support all of them. And probably the same for other tools. In my opinion, the datacontract CLI should expect a completely set up data quality tool and just run the configured checks on it - in this case, copy the checks and run them in Soda as you already implemented it. Please correct me, if you think I'm on the wrong path here.

Yes, I (of course) fully agree with that: the `datacontract` CLI should support many tools and should be as much tool-agnostic as possible, (at least) when it comes to quality checks.

On the other hand, most of the quality checker tools need some sort of configuration in order to properly work. And since those tools are wrapped within/executed by the `datacontract` CLI, the `datacontract` CLI has to "be aware" of that local configuration. To be more specific:
* The SodaCL CLI needs to be passed a configuration file (with the `-c` parameter), namely `soda-conf.yml` here: `soda scan -d duckdb_local -c soda-conf.yml contracts/data-contract-flight-route-quality.yaml`.
* And the underlying database, namely `duckdb_local` here, must be provided with the `-d` parameter, and match the database parameter specified in the SodaCL configuration section.

The `datacontract` CLI has to "be aware" of those parameters/configuration elements, so as to pass them to the underlying quality checker tool (SodaCL here).

main.go
da115115: > Would it help to allow multiple quality specifications in the data contract specification?

Yes, it would. When I presented internally (to colleagues) the working flow with data contracts (from the YAML specification in Git to the quality checks with Soda), almost the first thing they asked was "How can we make sure that different users will want different quality standards?"

I then thought that we could specify, in the data contracts, various sets of quality checks, each with some sort of identifiers and potentially some weight/flag mechanism, in order to enable/disable some of those sets and/or to provide a weighted score for all the sets.
To be more specific, we would have something like:
```yaml
quality:
  - basic:
      weight: 0.2
      type: SodaCL
      specification: "$ref: contracts/data-contract-flight-route-quality-basic.yaml"
  - premium1:
      weight: 0.4
      type: SodaCL
      specification: "$ref: contracts/data-contract-flight-route-quality-premium1.yaml"
  - premium2:
      weight: 0.4
      type: SodaCL
      specification: "$ref: contracts/data-contract-flight-route-quality-premium2.yaml"
```

stefannegele: I have opened an issue in the specification repo to address this. We can discuss this there.
https://github.com/datacontract/datacontract-specification/issues/12

main.go
stefannegele: > > I'm not sure what this applies to? https://github.com/datacontract/cli/pull/3#discussion_r1357876873?
> 
> I was referring to the following comment:
> 
> > Can this be integrated into `test / quality-check`?
> 
> Indeed, I thought you were meaning, with the slash (`/`) character, some sort of levels in the commands, as the slash (`/`) is usually the separator in file-system folder structures. I now realize that the (`/`) character was here meaning "or" and/or _aka_ (also known as): `test` or/aka `quality-check`. So, it is all clear by now, thanks!

Sorry for the confusion. We should drop `quality-check` and just go with `test`. The slash was meant as an "or" ;)

stefannegele: > > [...] I'm actually not sure, if the CLI should be responsible for setting up the test environment. Since Soda supports several other data sources next to DuckDB, we would need to aim to support all of them. And probably the same for other tools. In my opinion, the datacontract CLI should expect a completely set up data quality tool and just run the configured checks on it - in this case, copy the checks and run them in Soda as you already implemented it. Please correct me, if you think I'm on the wrong path here.
> 
> Yes, I (of course) fully agree with that: the `datacontract` CLI should support many tools and should be as much tool-agnostic as possible, (at least) when it comes to quality checks.
> 
> On the other hand, most of the quality checker tools need some sort of configuration in order to properly work. And since those tools are wrapped within/executed by the `datacontract` CLI, the `datacontract` CLI has to "be aware" of that local configuration. To be more specific:
> 
>     * The SodaCL CLI needs to be passed a configuration file (with the `-c` parameter), namely `soda-conf.yml` here: `soda scan -d duckdb_local -c soda-conf.yml contracts/data-contract-flight-route-quality.yaml`.
> 
>     * And the underlying database, namely `duckdb_local` here, must be provided with the `-d` parameter, and match the database parameter specified in the SodaCL configuration section.
> 
> 
> The `datacontract` CLI has to "be aware" of those parameters/configuration elements, so as to pass them to the underlying quality checker tool (SodaCL here).
> 
> A way to accommodate that need is to have an optional `configuration` parameter in the `quality` section of the data contract specification. It would give something like:
> 
> ```yaml
> quality:
>   type: SodaCL
>   configuration: "$ref: quality/soda-conf.yaml"
>   specification: "$ref: contracts/data-contract-flight-route-quality.yaml"
> ```
> 
> In the initialization step (when there is one) of the quality checks, a check would be made on the configuration file; if the configuration file exists (and seems valid?), nothing more needs to be done during the initialization step; otherwise, the configuration file would be generated as it is currently done by the [`sodaQualityInit()` function](https://github.com/data-engineering-helpers/datacontract-cli/blob/main/sodaQuality.go#L23).
> 
> Would that work for you?

This configuration is specific to the test execution, right? 

If not, how about using the [server object](https://datacontract.com/#server-object)? 

If yes, what about using a generic flag called "test-arguments", or something similar? E.g.:

```
datacontract test --test-arguments="-d duckdb_local -c soda-conf.yml" contracts/data-contract-flight-route-quality.yaml
```

da115115: Yes, the following would work:
```bash
datacontract test --test-arguments="-d duckdb_local -c quality/soda-conf.yml" contracts/data-contract-flight-route-quality.yaml
```

da115115: > If not, how about using the [server object](https://datacontract.com/#server-object)?

The server seems rather related to where the data is stored (_e.g._, S3, GCS, ADLS). But it makes think that it could also address other parts of the data stack as well, such as ingestion, data processing, data servicing (visualization, API), monitoring/observability, access control, catalog, metadata, etc. Moreover, a given data product could source several data sets, each one stored in/serviced by different technologies.

stefannegele: > > If not, how about using the [server object](https://datacontract.com/#server-object)?
> 
> By the way, the server seems rather related to where the data is stored (_e.g._, S3, GCS, ADLS). But it makes think that it could also address other parts of the data stack as well, such as ingestion, data processing, data servicing (visualization, API), monitoring/observability, access control, catalog, metadata, etc. Moreover, a given data product could source several data sets, each one stored in/serviced by different technologies.
> 

It's meant for the consumer to know where to get data. Similar to the Open API Server Object. https://learn.openapis.org/specification/servers.html

stefannegele: > Yes, the following would work:
> ```bash
> datacontract test --test-arguments="-d duckdb_local -c quality/soda-conf.yml" contracts/data-contract-flight-route-quality.yaml
> ```
> 
> Will you add that command option/parameter?
> 

Then let's go with it for now. I think I would be best if you add it directly in your branch?

da115115: Here we are: [e31d018](https://github.com/datacontract/cli/pull/3/commits/e31d018fb1b43b33349ac93e15f45b9029d5d0b9)

Example of the new command line for quality checks with options for SodaCL:
```bash
$ ./datacontract test --test-options "-d duckdb_local -c quality/soda-conf.yml" --file contracts/data-contract-flight-route.yaml --quality-file contracts/data-contract-flight-route-quality.yaml
```

stefannegele: Well done :)

da115115: And will you merge that pull request, then? :)

stefannegele: Didn't realize that it's final yet, because the title still starts with WIP. I will take another look soon.

da115115: Actually, there are plenty of things still to be done. But I would rather merge the pull request as is now (as it is working, even though in a somewhat limited way), while not closing it, so that I/we can continue to build upon it.

da115115: Sure!
Don't they already work today? Do you have some specific test failure in mind? The scenario I tested worked, but I haven't added specific tests (as I'm not good enough in Go yet). We should of course add some more tests.

stefannegele: Yes, tests would be great as well, especially integration tests in this case.

You can see what's failing in the checks section below:
https://github.com/datacontract/cli/actions/runs/6536887208/job/18055317649?pr=3

da115115: The tests have been fixed. Thanks!

stefannegele:
I plan to merge tomorrow when I have time for my adjustments. Thank you!

stefannegele (approved):

stefannegele:
Hi @da115115 
Just merged your PR and added some tickets. If you plan to work on one of them, leve a comment there, please.
https://github.com/datacontract/cli/issues/12
https://github.com/datacontract/cli/issues/13
https://github.com/datacontract/cli/issues/14

da115115:
> Hi @da115115 Just merged your PR and added some tickets. If you plan to work on one of them, leve a comment there, please. #12 #13 #14
> 
> Thanks for your work!

Sure, makes sense, will do!

PR #11, opened 2023-10-12
Split binary and lib

kesselborn:
this way, the functionality can be used as a lib as well


stefannegele:
Thanks, @kesselborn ! This is a quick and really useful addition.

PR #20, opened 2024-01-03
support local json schema

saugerDecathlon:
Hello, I am adding a support for Local Schema to be used by the lint command. I would need this feature for adding the Specification schema directly inside my project repository.

### DIFF
--- a/lint.go
+++ b/lint.go
@@ -10,7 +10,7 @@ import (
 	"os"
 )
 
-func GetSchema(location string) (*jsonschema.Schema, error) {
+func getSchema(location string) (*jsonschema.Schema, error) {
 	if IsURI(location) {
 		return getRemoteSchema(location)
 	} else {
@@ -50,7 +50,7 @@ func getRemoteSchema(schemaUrl string) (*jsonschema.Schema, error) {
 }
 
 func Lint(dataContractLocation string, schemaUrl string, target io.Writer) error {
-	schema, err := GetSchema(schemaUrl)
+	schema, err := getSchema(schemaUrl)
 	if err != nil {
 		return err
 	}
--- a/lint.go
+++ b/lint.go
@@ -7,20 +7,50 @@ import (
 	"github.com/qri-io/jsonschema"
 	"io"
 	"net/http"
+	"os"
 )
 
-func Lint(dataContractLocation string, schemaUrl string, target io.Writer) error {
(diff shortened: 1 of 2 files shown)


lint.go
```diff
@@ -7,20 +7,50 @@ import (
 	"github.com/qri-io/jsonschema"
 	"io"
 	"net/http"
+	"os"
 )
 
-func Lint(dataContractLocation string, schemaUrl string, target io.Writer) error {
+func GetSchema(location string) (*jsonschema.Schema, error) {
```
stefannegele, line 13: The function is only used here and therefore should be private (start with a lower case character, that's a Go convention). `getSchema`

It's extra important here because the term "schema" is ambiguous.

lint.go
```diff
@@ -7,20 +7,50 @@ import (
...
+	} else {
+		return getLocalSchema(location)
+	}
+}
+
+func getLocalSchema(schemaFileName string) (*jsonschema.Schema, error) {
+	if schemaFile, err := os.ReadFile(schemaFileName); err != nil {
+		return nil, fmt.Errorf("failed to read schema file: %w", err)
+	} else {
+        schema, err := parseSchema(schemaFile)
```
stefannegele, line 25: It looks like here's something wrong with the formatting / indent.

lint_test.go
```diff
@@ -20,6 +20,16 @@ func TestLint(t *testing.T) {
 			},
 			wantErr: false,
 			wantOutput: `🟢 data contract is valid!
+`,
+		},
+		{
+			name: "valid",
```
stefannegele, line 26: Please use more speaking test names (not like i did).

Maybe:

```go
name: "valid with remote file"
```

```go
name: "valid with local file"
```

stefannegele (changes requested):
Thanks again @saugerDecathlon for your contribution. I think that's very useful! 

I have found some minor things. If you resolve those i will merge immediatly. Do you need this in a release soon?

saugerDecathlon:
> Thanks again @saugerDecathlon for your contribution. I think that's very useful!
> 
> I have found some minor things. If you resolve those i will merge immediately. Do you need this in a release soon?

Thank you @stefannegele for the review. I have fixed the 3 issues. Yes, it would be cool to have it soon. It would avoid me to develop a work around for my project

stefannegele (approved):
Thanks, I will merge and release this in a patch today.

PR #40, opened 2024-02-13
Fix reading file from local server

Donutellko:
I've faced the following error when trying to execute `datacontract test datacontract.yaml` with a local fileserver. 
```
  File "/usr/local/anaconda3/lib/python3.11/site-packages/datacontract/engines/fastjsonschema/check_jsonschema.py", line 49, in read_json_file
    yield json.loads(file)
          ^^^^^^^^^^^^^^^^

TypeError: the JSON object must be str, bytes or bytearray, not TextIOWrapper
```

Configuration in datacontract.yaml:
```
servers:
  server_local_test:
    type: local
    path: "myfile.json"
    format: json
```

### DIFF
--- a/datacontract/engines/fastjsonschema/check_jsonschema.py
+++ b/datacontract/engines/fastjsonschema/check_jsonschema.py
@@ -46,7 +46,7 @@ def read_json_array(file):
 
 
 def read_json_file(file):
-    yield json.loads(file)
+    yield json.loads(file.read())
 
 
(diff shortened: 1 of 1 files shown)


simonharrer:
Thanks for reporting. You uncovered an even greater bug. I'll merge your PR and fix the rest that you uncovered. Basically, a few methods were used in two different, incompatible ways. :-)

PR #71, opened 2024-03-05
Added some more linters.

jeeger:
Added linter ideas from #4:
- Linter that checks that quality checks use existing field names for the model (currently only for SodaCL checks)
- Linter that checks that `noticePeriod` matches the ISO8601 duration format
- Linter that checks that all primary fields are also `unique` and `required`
- Linter that checks that all fields referenced from other fields references exist
- Linter that checks regex patterns for correctness


datacontract/data_contract.py
```diff
@@ -67,8 +80,9 @@ def lint(self) -> Run:
                 result="passed",
                 name="Data contract is syntactically valid",
                 engine="datacontract"
-            ))
-            run.checks.extend(ExampleModelLinter().lint(data_contract))
+                ))
+            for linter in self.enabled_linters:
+                run.checks.extend(linter.lint(data_contract))
```
simonharrer, line 85: When a linter is broken, the whole linting process stops. We could have an `except` clause in here to cater for this.

jeeger: Fixed.

PR #88, opened 2024-03-13
70 import format avro

ismailova-n:
Add importer for avro schemas


datacontract/data_contract.py
```diff
@@ -312,6 +313,11 @@ def _get_examples_server(self, data_contract, run, tmp_dir):
...
-        data_contract_specification = import_sql(data_contract_specification, format, source)
+        if format == "sql":
+            data_contract_specification = import_sql(data_contract_specification, format, source)
+        elif format == "avro":
+            data_contract_specification = import_avro(data_contract_specification, source)
+        else:
+            print(f"Import format {format} not supported.")
 
-        return data_contract_specification
+        return data_contract_specification
```
jeeger, line 323: Some leftover trailing whitespace

datacontract/imports/avro_importer.py
```diff
@@ -0,0 +1,63 @@
...
+            imported_fields[field.name] = Field()
+            imported_fields[field.name].type = map_type_from_avro(field.type.type)
+            imported_fields[field.name].description = field.doc
+    return imported_fields
+
+
+def map_type_from_avro(avro_type_str: str):
+    # TODO: ambiguous mapping in the export
+    if avro_type_str is None:
+        return None
```
jeeger, line 45: You should probably throw an exception or use an assert here. 
An empty type isn't allowed by Avro, and the schema import will probably throw an exception already if the type field is empty.
No need to turn this into a more confusing error condition (i.e. when the user of the data finds the None).

datacontract/imports/avro_importer.py
```diff
@@ -0,0 +1,63 @@
+import avro.schema
+
+from datacontract.model.data_contract_specification import \
+    DataContractSpecification, Model, Field
+
+
+def import_avro(data_contract_specification: DataContractSpecification, source: str):
```
jeeger, line 7: You can also annotate the return type (`-> DataContractSpecification`).

datacontract/imports/avro_importer.py
```diff
@@ -0,0 +1,63 @@
...
+    elif avro_type_str == "int":
+        return "int"
+    elif avro_type_str == "long":
+        return "long"
+    elif avro_type_str == "boolean":
+        return "boolean"
+    elif avro_type_str == "array":
+        return "array"
+    else:
+        return "variant"
```
jeeger, line 63: I don't think variant is a defined data contract data type, at least not in the specification. I would just throw an exception here.

datacontract/imports/avro_importer.py
```diff
@@ -0,0 +1,63 @@
...
+        return "binary"
+    elif avro_type_str == "double":
+        return "double"
+    elif avro_type_str == "int":
+        return "int"
+    elif avro_type_str == "long":
+        return "long"
+    elif avro_type_str == "boolean":
+        return "boolean"
+    elif avro_type_str == "array":
```
jeeger, line 60: Arrays need to be treated similarly to records, in `import_record_fields`, with `imported_fields[field_name].items = import_record_fields(field.items)`.

tests/test_import_avro.py
```diff
@@ -0,0 +1,60 @@
...
+          city:
+            type: string
+          state:
+            type: string
+          zipcode:
+            type: long
+    '''
+    print("Result:\n", result.to_yaml())
+    assert yaml.safe_load(result.to_yaml()) == yaml.safe_load(expected)
+    assert DataContract(data_contract_str=expected).lint(enabled_linters=set()).has_passed()
```
jeeger, line 60: Maybe add a test case for arrays and nested records as well.

jeeger (changes requested):

datacontract/imports/avro_importer.py
ismailova-n: https://github.com/datacontract/cli/issues/94

ismailova-n: Need some intro/ thought and examples on arrays in data contract

PR #92, opened 2024-03-14
Add great expectation option for the export command

saugerDecathlon:
Hello, 

I create this PR to add the great expectation option for the export command. 
The export generates a dictionary with a great expectation suite for each model in the contract. 
The export generates the list of expectations based on :
- the information in the Model definition with a hardcoded mapping 
- the list of expectation filled in the quality field for each model

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ## [Unreleased]
+- Added export format **great-expectations**: `datacontract export --format great-expectations`  
 
 
 
--- a/README.md
+++ b/README.md
@@ -440,22 +440,34 @@ datacontract export --format dbt
 
 Available export options:
 
-| Type               | Description                                             | Status |
-|--------------------|---------------------------------------------------------|--------|
-| `jsonschema`       | Export to JSON Schema                                   | ✅      | 
-| `odcs`             | Export to Open Data Contract Standard (ODCS)            | ✅      | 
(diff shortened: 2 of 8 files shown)


simonharrer:
Great contribution.

Can you add the following:
- Export format in the README
- Add entry in the CHANGELOG for the new export format
- The DataContract.export method should return an already serialized json.
- I'd expect to have the ID and Version of the datacontract somewhere in the great expectation suite. The idea is that if you look at the export, you know it's from a particular contract. You probably know best where this could be added/encoded.
- The result of the export should be a valid great expectation format. I'm not sure whether the current result is a valid one. We might need to make it similar to the export function for the jsonschema, where one has to pass in `--model orders` to get the jsonschema for a single model. Not sure, however.
- Some of the expectations are probably server-type specific. Think of the type mapping, which might result in different type checks. I don't expect this for this PR, but maybe for the next iteration. We currently have a type mapping construct for SQL server types, and a little bit for soda checks.

saugerDecathlon:
> Great contribution.
> 
> Can you add the following:
> 
> * Export format in the README
> * Add entry in the CHANGELOG for the new export format
> * The DataContract.export method should return an already serialized json.
> * I'd expect to have the ID and Version of the datacontract somewhere in the great expectation suite. The idea is that if you look at the export, you know it's from a particular contract. You probably know best where this could be added/encoded.
> * The result of the export should be a valid great expectation format. I'm not sure whether the current result is a valid one. We might need to make it similar to the export function for the jsonschema, where one has to pass in `--model orders` to get the jsonschema for a single model. Not sure, however.
> * Some of the expectations are probably server-type specific. Think of the type mapping, which might result in different type checks. I don't expect this for this PR, but maybe for the next iteration. We currently have a type mapping construct for SQL server types, and a little bit for soda checks.

PR's updated 😃

Issue #109, opened 2024-03-21
datacontract export --format pydantic


simonharrer:
Export to pydantic model to integrate with python pipeline

PR #117, opened 2024-03-25
First implementation of pydantic model export.

jeeger:
Implements Pydantic model export (#109). Currently missing export of description.
Also replaced uses of print() in `cli.py` with `console.print` to not interpret markup on export.


jochenchrist:
LGTM

PR #120, opened 2024-03-28
112 union with null (optional fields)

ismailova-n:
In avro an optional field is represented by a field type union of "null" and the actual type. Such fields are now imported as "required=false" fields in the datacontract, while all other fields are imported as "required=true"


jochenchrist:
Thanks for contributing

PR #143, opened 2024-04-16
added docker-compose.yml and documentation

enriquecatala:
A docker-compose.yml with basic specification that hopefully will help others to create the base image locally quicker with just a docker compose build command and being able to execute a docker compose run without the need of writing -v... every time


jochenchrist:
Thank you for your contribution!

simonharrer:
Can you clarify why the existing docker image does not suffice and why you need docker compose here? I‘m a little bit confused.

PR #151, opened 2024-04-23
feat: added preliminary support for reading delta lake tables from S3

### LINKED ISSUE #24: Adding S3 delta support
Is there a timeline / estimate for when S3 Delta or Databricks Server Objects might be available?

e.g. 
```
servers:
  production:
   type: s3
   location: s3://...
   format: delta
   delimiter: new_line
```

This tool looks like it has a lot of potential.
Thanks for all the great work so far.

chgl:
Closes #24

Still somewhat preliminary since it doesn't yet support Azure storage. Documentation is still missing, which I'd be happy to add if the changes look good!

To make the lint step pass I had to re-use the `fixtures/lint/custom_datacontract.schema.json` fixture with `delta` added as a format. Not sure if there's a better way until it's part of the official JSON schema.

Also not 100% happy about the duplication in `storage_options` and `setup_s3_connection`.

Let me know what you think!


jochenchrist:
Definitively looks good. 
Happy to merge, as soon as you update the README.

jochenchrist:
Also, I updated the official Data Contract Specification to allow `delta` as format

tests/test_test_s3_delta.py
```diff
@@ -0,0 +1,58 @@
...
+    ) as minio_container:
+        yield minio_container
+
+
+def test_test_s3_delta(minio_container, monkeypatch):
+    monkeypatch.setenv("DATACONTRACT_S3_ACCESS_KEY_ID", s3_access_key)
+    monkeypatch.setenv("DATACONTRACT_S3_SECRET_ACCESS_KEY", s3_secret_access_key)
+    data_contract_str = _prepare_s3_files(minio_container)
+    data_contract = DataContract(
+        data_contract_str=data_contract_str, schema_location="fixtures/lint/custom_datacontract.schema.json"
```
jochenchrist, line 34: you should be able to remove the schema_location, now

chgl: Great, thank you! Seems to work now.

PR #152, opened 2024-04-23
feat(storage): implements abstraction for storing data contracts

PatrickfBraz:
## Context

The CLI tool enables a series of iterations and management of Data Contracts. However, in a corporate environment these contracts will not be managed in a local environment. Thus, we had an idea of ​​creating an abstraction to allow data contracts to be managed from environments other than the local CLI execution environment.

### Makefile

Just a simple addition to make life easier for those who like to use Makefile.

Typing the following terminal command:
```shel
make help
```
Produces:
image

### Storage abstraction

The abstraction was designed and created to allow flexibility for the development of new integrations with storage such as Cloud Storage, s3, and others.

### Settings with Pydantic

Just a way to facilitate access to configurations via environment variables in a reliable and validated way. But possibly it can be thought of in another way.

See: https://docs.pydantic.dev/latest/concepts/pydantic_settings/

### Evidence

Running the CLI with storage type as 'google_cloud': 
```shell
datacontract init --overwrite
```
Produces:
image

### Considerations

- There are some changes that were made only by the code formatter, but I decided to upload them anyway to adjust the unformatted files.
- Some unit tests were added for the new storages, but they are still very simple.


jochenchrist:
@PatrickfBraz thank you for this PR.
To minimize the format diffs, could you merge the latest main branch and run `ruff format`?

simonharrer:
The CLI should work locally but it can resolve files via URL. You can add a resolver that resolves appropriately for google cloud storage just like we have a resolver for http and http at the data mesh manager. But supporting a complete new file system including write and delete support is the wrong way to go here I think.

PatrickfBraz:
@jochenchrist @simonharrer 
Thanks for the feedback! Well, my idea with including this abstraction layer was to enable the CLI to execute commands on storage layers other than local storage. This opens up a series of possibilities for the application, at least from my point of view. I currently don't have any service that uses the CLI in production, but I have some visions of how I can use it in the future in the company I work for. Anyway, if that wasn't the intention of the tool, could I have access to any of your roadmaps and help with the tool's evolution? I believe it has great potential.

simonharrer:
We currently focus on adding more imports and exports (or improve the existing ones), as well as support testing the contract against various server technologies. Here, we really need help! :-) And that's basically the roadmap. Does this help?

PatrickfBraz:
It helps too much. I'll study the code a little better and see how I can help with this. Do you believe that this PR still has potential or should we close it?

PatrickfBraz:
I agree that maybe this really is the best way.

PR #154, opened 2024-04-24
feat(import): adds bigquery importer

PatrickfBraz:
## What is :new:

* Abstraction to create Data Contracts import classes
* New class to import BigQuery tables

## :warning: Considerations

* The new class 'STILL' does not support importing multiple tables.
* It may be that over time the addition of new import classes makes the code slower due to dynamic import.

## :red_circle: Breaking Changes

* The 'import' command now receives dynamic user-supplied prompts. The '--source' parameter no longer exists.

## Operation

Testing with the following file:
image

The file was created in the root of the repository. Running the following command:
```shell
datacontract import --format sql
```
The following screen is rendered:
image

Then:
image

### Testing a BigQuery table

Running the following command:
```shell
datacontract import --format bigquery
```
The following screen is rendered:
image

The parameters passed were:
image

Then:
image

### How to reproduce:

If not configured, configure an environment variable `GOOGLE_APPLICATION_CREDENTIALS` with the path to a service account that has the necessary permissions and then enter the full name of an existing table in one of your projects.

## Tests
image


PatrickfBraz:
@simonharrer 
It may be interesting to discuss how this import can be done, as I believe that both tables and views, but also stored procedures can be contained within a Data Contract. Any thoughts on improvements here?

simonharrer:
Thanks for this PR. I really appreciate the effort you take.

- Import-wise, I'm missing a type mapping of the BigQuery types to the Data Contract Specification types. Have a look at what we are doing for Avro and SQL. This is, however, crucial, as the result of the import should be a valid Data Contract, and without a type mapping, it isn't.
- Refactoring-wise, I do not approve. Let's try to keep the implementation very simple. If you need import-specific CLI parameters, simply prefix the parameters with "bigquery-" and pass them on. You can have a look at how we're implementing it for the exports.
- The logic you have implemented connects to BigQuery, gets the description of a table via their API, and then converts this to a datacontract. The GET call to the API is somewhat glue code, and the actual import uses the JSON result object of that API call. We could also think of starting with this table result format, and keep the "GET" call to the API out of scope as it can be easily done using existing google tools to get that JSON object downloaded, and then convert the json object into a data contract.
- The CLI is currently not interactive. We want to keep it that way for now. So every parameter that is required has to be passed in and not asked from the user interactively. We might change that later, but at the moment, let's keep the CLI tool simpler.
- Please keep a PR focused. This PR adds comments, does a refactoring of the structure of the import, and adds an import for BigQuery. Let's try to focus solely on the new feature here.

PatrickfBraz:
Well, I think I understand. I'm going to close this PR and open another one with a simpler implementation based on a JSON file.

PR #159, opened 2024-04-26
Add output path option for export command

### LINKED ISSUE #156: Add --output parameter to write to file
Issue: When piping stdout to a file, and when run inside Docker, there's a line-length-limit of 80 chars (can probably be overriden somehow).

Hence generated HTML tend to break if these extra line-breaks appear in the middle of e.g. css strings.

Suggested solution: Introduce a --output parameter that uses python's file writing to circumvent these environment specific issues.

Impact: Allows generating HTML as a github action, which is a containerized workflow.

jochenchrist:
closes #156

### DIFF
--- a/datacontract/cli.py
+++ b/datacontract/cli.py
@@ -3,17 +3,16 @@ from importlib import metadata
 from pathlib import Path
 from typing import Iterable, Optional
 
-
 import typer
 from click import Context
-
 from rich import box
 from rich.console import Console
 from rich.table import Table
 from typer.core import TyperGroup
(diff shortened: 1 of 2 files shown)


jochenchrist:
Option just `output` or `output-path` or something?.
Do we need it in the Class es well?

simonharrer:
We could offer it for import (datacontract) and test (json output, although we need a —format=json option here too) as well. Output sounds fine I suppose.

jochenchrist:
Merging, let's add new issues for the other options, if needed.

PR #176, opened 2024-05-06
Bugfix of importing avro schema that includes an array inside a nested record or object

enriquecatala:
I found a bug when trying to import one of my complex avro schema definitions that includes among other things, some arrays inside nested fields with both objects and records. I destillated a little bit the sample and created a fix with the corresponding unit test.


jochenchrist:
Thank your for your contribution and fixing this bug.

Please keep the PR focused. May I ask you to revert the changes for 
- conda_environment
- .vscode/
- README.md

Then I'd be happy to merge.

For dev setup (conda, vs code, docker compose...),  please create an issue first to discuss, as these settings need maintanance as well, which we should discuss and evaluate benefits and long term costs.

jochenchrist:
Thanks for your contribution!

PR #180, opened 2024-05-07
fix(export/jsonschema_converter.py): Add more keys to to_jsonschema()

nyeongna:
**datacontract/export/jsonschema_converter.py**
- it now supports **_pattern, enum, minLength, maxLength, and_** [**_patternProperties_**](https://json-schema.org/understanding-json-schema/reference/object#patternProperties) when exported to jsonschema
- since data_contract_specification doesn't have the key "patternProperties" in Field Object, I added an logic to check if patternProperties is used or not. Please check out the details in the code.

**tests/fixtures/local-json-complex/***
- added test code that can check more complex model (which has enum, minLength, maxLengh, patternProperties as example)

**tests/test_export_jsonschema.py**
- edited "expected json schema" as it conflicted with the newer jsonschema_converter.


jochenchrist:
LGTM.
Thanks for contributing.

Issue #110, opened 2024-03-22
datacontract import --format bigquery


simonharrer:
Import from BigQuery tables object: https://cloud.google.com/bigquery/docs/reference/rest/v2/tables#resource:-table

It can also be thought about importing from BigQuery Schema.

jochenchrist:
This issue is stale.
Anyone interested in this feature?

jochenchrist:
Closing issue as stale. Please reopen, if you need this feature.

jochenchrist:
Multiple ideas how the sources can come from:

- BigQuery tables object (https://cloud.google.com/bigquery/docs/reference/rest/v2/tables#resource:-table)
- SQL DDL as generated in the console (?)
- Direct connection to BigQuery to read schema

We should start with one option

simonharrer:
It could also be the BigQuery schema language that uses JSON.

PatrickfBraz:
It's definitely a great idea :100: 

What I think:

- Import from JSON (which follows the format of the [Table object](https://cloud.google.com/bigquery/docs/reference/rest/v2/tables#Table)) in which the path is passed by the user requires least effort to implement. However, it requires the user to make the API call and generate the JSON file to be consumed by the CLI.

- Import from json that will be fetched by the CLI itself. This requires the user to provide service account information so that we can authenticate the call to the Google server. This alternative adds the complexity of managing authentication on the server.

Starting from the premise of simplicity, we can start with an MVP with the first option and if the adoption is interesting, we can basically take advantage of everything that was implemented and release a version 2 that does the JSON search.

What do you think?

PatrickfBraz:
@jochenchrist 
Yes! You can count on me! I'll soon raise the PR with the contribution!

jpraetorius:
@PatrickfBraz Apologies then – I've had some time today and this one got pointed out as an urgent one, so I took a stab at it. Would prefer if you'd have a look, as I'm sure that you've thought about things a bit deeper than I did.

I also had a look and given the dependencies we already have it shouldn't be to terribly complicated to read tables directly of Bigquery directly, I've just left that as a next step, as it also means a couple more changes to the CLI since we need more parameters and probably want to make `source` optional

jpraetorius:
I have one more change pending that would also introduce said 'direct import from the API' – you want this under this ticket too, or do I create a new one?

jochenchrist:
Resolved with #197

PR #194, opened 2024-05-13
Issue/110 import from bigquery

jpraetorius:
This should allow to import from a Bigquery JSON File as e.g. `bq show --format prettyjson <table-id>` produces it


jpraetorius:
Addresses #110

jochenchrist:
LGTM

Issue #187, opened 2024-05-10
Export model to class definition


pixie79:
It would be great to be able to export the model definition to different language definitions.

Eg. Python Class & Go Structs

jochenchrist:
Code generation, including generating classes, such a Python, Java, Go, ... from the model is the idea of the export feature.

For Python, there is support to export to Pydantic classes

```
datacontract export --format pydantic-model  https://datacontract.com/examples/orders-latest/datacontract.yaml
```

We are open to supporting more languages based on use-cases, needs and contributions.

PR #195, opened 2024-05-14
Add ability to export to go types

pixie79:
#187 

Add the ability to export *go* structs.

This will need extension once the handling of logicalTypes is added.


jochenchrist:
Thanks for your contribution.
Please add a pytest for this export :)
Also, please run 
`ruff format` to format the code

pixie79:
I have added a test, ideally I will add some more tests to handle all the logical types once it has been decided by the project how these are being handled. At which point I think the test datacontract.yml needs to be more extensive.

jochenchrist:
LGTM.

Only think that surprises me a bit are the json & avro tags. 
What is the rationale behind them?

jochenchrist (approved):

pixie79:
The reason for the tags is it helps with serialisation of data to/from json. We already know that information from the data contract specification so it helps to add it here. If it is not use by the end user later if does no harm being there, but if they need it then it saves a lot of work :)

jochenchrist:
OK, agree with JSON. Still not quite sure, if Avro is commonly used with Go, but let's see if we get some feedback here. Happy to merge.

PR #197, opened 2024-05-14
Issue/110 part2 read from bigtable api

jpraetorius:
Another thing for #110 

This time we add the capability to directly query from the API, so people don't have to export JSON Files in order to import an existing schema. This assumes they can authenticate towards the API in one of the ways the Google Python Client Library prescribes.

### DIFF
--- a/README.md
+++ b/README.md
@@ -193,16 +193,16 @@ Commands
  Download a datacontract.yaml template and write it to file.                                    
                                                                                                 
 ╭─ Arguments ──────────────────────────────────────────────────────────────────────────────────╮
-│   location      [LOCATION]  The location (url or path) of the data contract yaml to create.  │
-│                             [default: datacontract.yaml]                                     │
+│   location      [LOCATION]  The location (url or path) of the data contract yaml to create. │
+│                             [default: datacontract.yaml]                                    │
 ╰──────────────────────────────────────────────────────────────────────────────────────────────╯
 ╭─ Options ────────────────────────────────────────────────────────────────────────────────────╮
-│ --template                       TEXT  URL of a template or data contract                    │
-│                                        [default:                                             │
-│                                        https://datacontract.com/datacontract.init.yaml]      │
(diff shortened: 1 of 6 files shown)


README.md
```diff
@@ -651,24 +651,37 @@ data products, find the true domain owner of a field attribute)
...
-│ *  --source        TEXT             The path to the file or Glue Database that should be imported. [default: None] [required]   │
-│    --help                           Show this message and exit.                                                                 │
-╰──────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────╯
+╭─ Options ──────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────╮
+│ *  --format             [sql|avro|glue|bigquery]  The format of the source file. [default: None] [required]                                                                            │
+│ *  --table              TEXT                      List of table ids to import from the bigquery API (repeat for multiple table ids). [default: None] [required]                        │
+│ *  --bt-project-id        TEXT                      The id of the project that we should query on bigtable. [default: None] [required]                                                 │
```
simonharrer, line 657: I'd use `--bigquery-project`, `--bigquery-dataset`, and `--bigquery-table` as the parameters. And in the description, we should not use bigtable, but bigquery instead. We should make `--bigquery-table` optional and import all tables within the `--bigquery-dataset` when it is missing.

simonharrer (changes requested):

README.md
simonharrer: Because of how the CLI help works, having a prefix similar to the format helps to group the parameters together that are specific to the import format. In the future, we will probably move towards separate importers, with their very own set of parameters. With this, we can get rid of these prefixes at some point. But this is not too relevant right now.

jpraetorius: Just to follow up on it: I deliberately named `table` without a prefix, as I imagined it might come in handy for others importers as well, assuming that other formats also allow to have multiple models. But I guess we can change things when that comes around

simonharrer (approved):

PR #200, opened 2024-05-15
feat(export/jsonschema): supports array type

nyeongna:
Thanks for merging the previous export/jsonschema PR.
I added a bit of code so that it can supports "array" type. (adjusted test code accordingly as well).
Have a look and give some comments if any. Thanks.


jochenchrist:
Thanks for contributing!

Issue #91, opened 2024-03-14
import --format jsonschema


simonharrer:

markusklemens:
I'm very interested in this feature... any plans to have this added in the near future?

simonharrer:
It's planned, but I cannot promise any deadline here. We'd love to have open source contributions, though. Perhaps something you might want to step up to implement? The export is already there. :-)

PR #201, opened 2024-05-15
91 JSON Schema

pixie79:
Add JSON Schema Import function and extend export.

Currently import does not handle $ref (not sure what we want to do there)

Export does not export definitions (could not see where to extend that bit)


pixie79:
This is to fix issue #91

simonharrer:
Thanks for this PR! Looks fine, but let's exclude the definitions here for now as this needs discussion.

1. The reference mechanism of JSON schema differs from Data Contract Specification.
2. The Data Contract Specification currently does not support the allOf/oneOf/anyOf logic of JSON schema.

Let's discuss each of those separately.

Regarding 1: Without the allOf/oneOf/anyOf logic, we could implement an import that makes use of the definitions in DCS. We could use the $ref mechanism to link to the referenced definition. No problem here. The difference, however, is that one can override the fields (in JSON schema, you cannot). But this should not be an issue here, but more an issue when you'd do an export to JSON schema when using $ref.

Regarding 2: We could support this with `config` options, e.g., like this:
```yaml
models:
  my-model:
    config:
      jsonSchemaAllOf:
        - ref1
        - ref2
```

pixie79:
Agreed, I just mentioned it as something that needs to be handled in the future so that it did not get lost. Happy to leave as is for now while a debate happens as to the best options.

simonharrer:
Sure, thanks for the great PR. I've created an issue to track the discussion around the discovered issues.

Issue #193, opened 2024-05-13
HTML Export: also export examples


simonharrer:

jochenchrist:
resolved with #193

PR #203, opened 2024-05-15
Issues/193 fix all todos in html export

jpraetorius:
This not only addresses the Examples mentioned in #193 but additionaly

- exports Definitions
- adds Examples of fields in the Models section (code was there but never did anything)
- exports nested fields of fields in the Models section. This currently only goes one level deep, as it already is a proper pain without partials ;)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - `datacontract import --format jsonschema`: Import from JSON schema (#91)
 - `datacontract export --format jsonschema`: Improved export by exporting more additional information
 - `datacontract publish`: Publish the data contract to the Data Mesh Manager
+- HTML Export: Added support for Service Levels, Definitions, Examples and nested Fields (one level deep)
 
 ## [0.10.3] - 2024-05-05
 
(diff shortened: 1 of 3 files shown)


jpraetorius:
This is how things look like

SCR-20240515-pbks

simonharrer:
jinja partials would be awesome

simonharrer:
layout of the flags definition seems broken

jpraetorius:
> layout of the flags definition seems broken

That was a simple showcase to test out how `enum` would render. It doesn't have any other properties defined that would go in the third column.
This is how it would be with some more content (but we can also try to move the enums to a different place)
SCR-20240516-idpx

jochenchrist:
I'd put them into the description column, as the enum list could become quite long.
To reduce abstraction, maybe I'd call them "Enum" instead of "Allowed values"?

jpraetorius:
So, here we go:
- introduced partials and theoretically fields now nest as deep as they are defined (I've only tried it two levels deep)
   - I'm not a 100% happy with how that is presented, some sort of collapse or different background shading would be good I think, but all of that is complicated to do with a table
- named enums enums and moved them into the description column. I've tried to make the labels stand out a bit more, so the contents of that column are a bit easier to parse
- moved all of the other blocks into partials, to keep the main template a bit more slim

jochenchrist:
The partials are a bit dangerous, as you need to remember to add them (as data files) to the MANIFEST.in

Issue #196, opened 2024-05-14
Support for MS Sqlserver 


RobertLD:
Our company is actively pursuing stricter guarantees on cross-team data via the data contract-cli. We currently make use of MS SqlServer and would like to know if this is supported within the data contract cli?

jochenchrist:
Currently, it is not implemented, but should not be too difficult to implement, as soda-core supports ms sqlserver.

As a precondition, we'd need to add sqlserver to the server specification
https://github.com/datacontract/datacontract-specification/?tab=readme-ov-file#server-object

jochenchrist:
Would you (or your team) have some capacity to contribute a Pull-Request?

RobertLD:
I'm sure I can find some time to  pull the code and investigate, hopefully sometime this week or next.

jochenchrist:
Added `sqlserver` to data contract specification: https://github.com/datacontract/datacontract-specification/?tab=readme-ov-file#sql-server-server-object

RobertLD:
> Added `sqlserver` to data contract specification: https://github.com/datacontract/datacontract-specification/?tab=readme-ov-file#sql-server-server-object

Including the supporting code, or just preparing the doc?

simonharrer:
Jochen added the server type to the specification.

What is now missing is adding support in the CLI within the python code base. Do you want to have a try for this @RobertLD ? It should be straight forward mapping to soda sqlserver type.

RobertLD:
I'll take a crack at it once I get my bearings

RobertLD:
> Added `sqlserver` to data contract specification: https://github.com/datacontract/datacontract-specification/?tab=readme-ov-file#sql-server-server-object



> Jochen added the server type to the specification.
> 
> What is now missing is adding support in the CLI within the python code base. Do you want to have a try for this @RobertLD ? It should be straight forward mapping to soda sqlserver type.

I've published a rough PR for the changes because the specification published earlier for SQLServer doesn't match exactly what I would expect (namely missing the driver, etc.)

With some guidance or an update of the spec I can continue work on it
https://github.com/datacontract/datacontract-cli/pull/204


ty

RobertLD:
Completed in PR #204

PR #204, opened 2024-05-15
[196] Add support for MSSQL

RobertLD:
Add framework for SQLServer support.

Currently the specification differs from what is expected by the SODA-CLI so this will have to be updated to match before the added tests will pass 

https://docs.soda.io/soda/connect-mssql.html

Related to: #196


tests/fixtures/sqlserver/datacontract.yaml
```diff
@@ -0,0 +1,28 @@
...
+    driver: ODBC Driver 18 for SQL Server
+models:
+  my_table:
+    type: table
+    fields:
+      field_one:
+        type: varchar
+        required: true
+        unique: true
+        regex: "[A-Za-z]{2}-\\d{3}-[A-Za-z]{2}$"
```
simonharrer, line 23: should probably be pattern

datacontract/model/data_contract_specification.py
```diff
@@ -31,6 +31,10 @@ class Server(pyd.BaseModel):
     token: str = None  # Use ENV variable
     dataProductId: str = None
     outputPortId: str = None
+    trusted_connection: str = None
+    trust_server_certificate: str = None
+    encrypt_connection: str = None
+    driver: str = None # Use ENV variable
```
simonharrer, line 37: do you need those fields? could we get by without them?

RobertLD:
The default provided values are not entirely suitable for a large majority
of connections. Users that are using Kerberos or active directory, for
example require a trusted connection and will not provide a username or
password.

RobertLD:
It certainly can be. I'll make that change.

pyproject.toml
```diff
@@ -40,7 +40,8 @@ dependencies = [
   "opentelemetry-exporter-otlp-proto-http~=1.16.0",
   "deltalake~=0.17.0",
   "boto3>=1.34.41,<1.34.99",
-  "botocore>=1.34.41,<1.34.99"
+  "botocore>=1.34.41,<1.34.99",
+  "pymssql==2.3.0"
```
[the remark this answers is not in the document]
RobertLD: It certainly can be. I'll make that change.

datacontract/model/data_contract_specification.py
RobertLD: The default provided values are not entirely suitable for a large majority
of connections. Users that are using Kerberos or active directory, for
example require a trusted connection and will not provide a username or
password.

tests/fixtures/sqlserver/datacontract.yaml
RobertLD: Hey Simon, how do you mean?

simonharrer: regex is not a known field in the data contract spec, should be named pattern, I think.

datacontract/model/data_contract_specification.py
RobertLD: > do you need those fields? could we get by without them?

I am open to alternatives if you'd rather have them be defined as environmental variables, for example

tests/fixtures/sqlserver/datacontract.yaml
simonharrer: a then it is wrong there as well. Good catch! :-)

datacontract/model/data_contract_specification.py
jochenchrist: I'd go for environment variables and sensible defaults, as they probably would not be part of the datacontract.yaml

simonharrer: Hm, good question. I am no export of sqlserver, so I don't really know which way is better.

For me, env variables should be used for secrets or things that differ from user to user.

If, however, you need those fields in order to know how to connect to the data, and it can only be done knowing those infos, then we should add them to the spec. 

Does this help? What would you do with this guideline?

PS: I just noticed that the three fields are of type boolean, but represented as strings in the pydantic data contract spec model.

RobertLD: @simonharrer I wanted to match the current code style as much as possible and use string representation as that seemed to be the default.

The only one absolutely required without any reasonable defaults is MSSQL driver. The others can be made as env variables

jochenchrist:
@RobertLD There was an error in the Data Contract JSON Schema (https://datacontract.com/datacontract.schema.json). I fixed the list of required fields there.

RobertLD:
> @RobertLD There was an error in the Data Contract JSON Schema (https://datacontract.com/datacontract.schema.json). I fixed the list of required fields there.

Sorry to bother you again but there appears to be one more issue with the schema. SQLServer types aren't supported in the field types yet. There are a whole bunch of them, but for now datetime and datetime2 are missing.

image

jochenchrist:
> Sorry to bother you again but there appears to be one more issue with the schema. SQLServer types aren't supported in the field types yet. There are a whole bunch of them, but for now datetime and datetime2 are missing.

The `field.type` represents a logical type, and it is a finite list of types. https://datacontract.com/#data-types
You'd need to define a mapping to / from these types.

In addition, you can specify the physical MS SQL Server Types in a config block, like this: https://datacontract.com/#config-object

```
models:
  my_table:
    fields:
      my_field_1:
        description: Example for a timestamp
        type: timestamp
        config:
          sqlserverType: DATETIME2
```

This config object is quite new, if you have any improvement ideas, I'd be happy to hear them.

RobertLD:
> type: timestamp_ntz
>         config:
>           sqlserverType: DATETIME2

I updated the datacontract to use the config object and it doesn't appear to be actually overwriting the type to check against.

image

image

image

We're in the end-zone now; this is the last element before I formalize the work here and rebase to something cleaner. Thank you a lot for the support in doing the work!

simonharrer:
Thanks already for your dedication.

You would need something like this for the type mapping
https://github.com/datacontract/datacontract-cli/commit/b6bd8fbec07a0531698e6cf061a19c3c72bbe041

And we need a type mapping for sqlserver anyway in datacontract/export/sql_type_converter.py

RobertLD:
> Thanks already for your dedication.
> 
> You would need something like this for the type mapping [b6bd8fb](https://github.com/datacontract/datacontract-cli/commit/b6bd8fbec07a0531698e6cf061a19c3c72bbe041)
> 
> And we need a type mapping for sqlserver anyway in datacontract/export/sql_type_converter.py

Added the serializer for sqlserver data-types and added the correct driver to the docker file. All of the functionality ought to be there now.


I've also rebased ontop of main so it should be a bit easier to review now

datacontract/engines/soda/connections/sqlserver.py
```diff
@@ -0,0 +1,41 @@
...
+            "type": "sqlserver",
+            "host": server.host,
+            "port": str(server.port),
+            "username": os.getenv("DATACONTRACT_SQLSERVER_USERNAME", ''),
+            "password": os.getenv("DATACONTRACT_SQLSERVER_PASSWORD", ''),
+            "database": server.database,
+            "schema": server.schema_,
+            "trusted_connection": os.getenv("DATACONTRACT_TRUSTED_CONNECTION", False),
+            "trust_server_certificate": os.getenv("DATACONTRACT_TRUST_SERVER_CERTIFICATE", False),
+            "encrypt":  os.getenv("DATACONTRACT_ENCRYPTED_CONNECTION", True),
```
simonharrer, line 35: I'd prefix the other settings to DATACONTRACT_SQLSERVER_ as well

Dockerfile
```diff
@@ -22,7 +22,7 @@ RUN python -c "import duckdb; duckdb.connect().sql(\"INSTALL httpfs\");"
 
 FROM ubuntu:22.04 AS runner-image
 
-RUN apt-get update && apt-get install --no-install-recommends -y python3.11 python3.11-venv && \
+RUN apt-get update && apt-get install --no-install-recommends -y python3.11 python3.11-venv msodbcsql18 && \
```
simonharrer, line 25: Why is this necessary? Can you share a little bit on that?

simonharrer (changes requested):
Nice work. Only two minor things.

Additionally, please add an entry to CHANGELOG.md what you have done and make sure to add the server type in the README.md

Future work could be to support for `datacontract export --format sql` for sqlserver along with a test and make sure that the types are correctly set during `datacontract import --format sql`.

datacontract/engines/soda/connections/sqlserver.py
RobertLD: Fair enough; I'll make them more consistent

RobertLD:
> Nice work. Only two minor things.
> 
> Additionally, please add an entry to CHANGELOG.md what you have done and make sure to add the server type in the README.md
> 
> Future work could be to support for `datacontract export --format sql` for sqlserver along with a test and make sure that the types are correctly set during `datacontract import --format sql`.


Made the requested changes to the changelog and readme. As for the future work, I can put together two issues for those to stop this issue from expanding to gigantic proportions. I'll start working on those next; if you're alright with splitting them out into separate issues

Dockerfile
simonharrer: We do not use the Dockerfile for test execution. We run the tests with pytest locally, and we use github ci to execute them on CI. Testcontainers is only used to provision databases so we test against the actual thing. 

We should install that driver in the github workflow file instead. And we should disable the test when not in CI as it would otherwise force everybody to install those drivers on their machines to male the tests run.

simonharrer:
Sure, feel free to create those issues.

Dockerfile
RobertLD: These two statements seem to be conflicting. If you want to execute that test and have it succeed it requires the driver. As for the CI I can add it there as well, but as it stands I'm unsure what the consensus is for having it in the Dockerfile as well

simonharrer: I think both comments are valid. We need the lib in the Dockerfile for `datacontract test` command, but we also need the lib in the CI pipeline so we can run `pytest` in GitHub actions as well as run the test locally.

RobertLD: Added to both areas

RobertLD:
> My last issue is, that in order to run the tests, I now have to install the ODBC driver on my machine.
> 
> Instead, I'd like that this sqlserver test is ignored by default, and only executed in the CI/CD pipeline. Can you implement that? Otherwise, it is good to go. Thanks for your dedication here!

Oh course! I'll implement these changes

RobertLD:
> My last issue is, that in order to run the tests, I now have to install the ODBC driver on my machine.
> 
> Instead, I'd like that this sqlserver test is ignored by default, and only executed in the CI/CD pipeline. Can you implement that? Otherwise, it is good to go. Thanks for your dedication here!

Changes made; easier than anticipated haha

Dockerfile
RobertLD: Updated to skip the test outside of CI/CD and likewise removed from dockerfile

.vscode/settings.json
simonharrer, line 1: can you please remove this file?

Dockerfile
```diff
@@ -22,7 +22,7 @@ RUN python -c "import duckdb; duckdb.connect().sql(\"INSTALL httpfs\");"
 
 FROM ubuntu:22.04 AS runner-image
 
-RUN apt-get update && apt-get install --no-install-recommends -y python3.11 python3.11-venv msodbcsql18 && \
+RUN apt-get update && apt-get install --no-install-recommends -y python3.11 python3.11-venv && \
```
simonharrer: I'm confused. Why did you remove this here?

tests/test_test_sqlserver.py
```diff
@@ -23,7 +24,7 @@ def remove_container():
 
     request.addfinalizer(remove_container)
 
-
+@pytest.mark.skipif(not os.getenv('CI'), reason="Skipping test outside CI/CD environment")
```
simonharrer, line 27: Ah, simple enough. Thanks. :-)

.vscode/settings.json
RobertLD: Accidental inclusion, removed

Dockerfile
RobertLD: I guess I misunderstood you; I thought we didn't want it included in the dockerfile but only in the git workflow?

simonharrer: I'm fine with it being in the Dockerfile as we'll ship the CLI as a docker container, and it should come prepacked with necessary software packages that it works out of the box.

RobertLD:
I'll add it back in a few hours, then. My mistake.

simonharrer:
Thanks for your dedication to bring this PR through! :-)

RobertLD:
Of course, thank you for the assistance in getting it done. I'll follow up
with those smaller issues shortly and hopefully knock those out soon.

PR #212, opened 2024-05-21
Add py.typed marker for top-level package

janhicken:
When other applications depend on the Data Contract CLI Python library, in order for type checkers to incorporate the type annotations already present in the code base, a `py.typed` marker file needs to be present.

See [PEP 561](https://peps.python.org/pep-0561/#packaging-type-information) for more details.


simonharrer:
Makes sense. I'll merge this.

jochenchrist:
Not sure if it needs to be part of manifest.in

Issue #135, opened 2024-04-09
datacontract export --format dbml


jchurel:
Would be cool to have this export option to have a visual view via [dbdiagram.io ](https://dbdiagram.io/)

dbml == database markup language

simonharrer:
Sure, why not. I don't know what currently has more adoption (mermaid, dbml, plantuml, etc.) but I'm open to any contribution here. :-)

jochenchrist:
I think dbml is a valid choice. Besides dbdiagram.io (proprietary) there are some other tools that support it (e.g., https://github.com/softwaretechnik-berlin/dbml-renderer/)

@jchurel Would you be interested in contributing the export?

simonharrer:
Closing this as #215 implements this. :-)

simonharrer:
@jchurel could you have a look at how the feature was implemented, and whether it solves your use case? Would love to hear your feedback. :-)

jchurel:
@simonharrer of course, how can I make a test?
I've tried `datacontract export --format dbml https://datacontract.com/examples/orders-latest/datacontract.yaml > datacontract.dbml` but have this error:

```
╭─ Error ─────────────────────────────────────────────────────────────────────────────────────────────────────────────╮
│ Invalid value for '--format': 'dbml' is not one of 'jsonschema', 'pydantic-model', 'sodacl', 'dbt', 'dbt-sources',  │
│ 'dbt-staging-sql', 'odcs', 'rdf', 'avro', 'protobuf', 'great-expectations', 'terraform', 'avro-idl', 'sql',         │
│ 'sql-query', 'html', 'go', 'bigquery'.                                                                              │
╰─────────────────────────────────────────────────────────────────────────────────────────────────────────────────────╯```

simonharrer:
It is not yet released. You need the lastest unreleased dev version. You would need to install it from the source, see the development guide in the README. Or wait until our next release.

jchurel:
I've just tested `python3 cli.py export --format dbml  https://datacontract.com/examples/orders-latest/datacontract.yaml > test.dbml` 

There are errors concerning notes that are too long and are creating a new line: `Invalid newline encountered while parsing` 

You can test the output here to be sure: https://dbdiagram.io/

Except that, once corrected, the dbml looks fine.

jpraetorius:
Happy to – I've been running my gerenrated code through the tool already, as the spec isn't really concise about such things.

jpraetorius:
[...]

PR #215, opened 2024-05-21
Add Support to Export to DBML

jpraetorius:
- Create basic project info from what the datacontract has
- Map all the Models into tables, _not_ taking into regard nested fields (as there is no way to express them anyway)
- Also create references, when they are given for a field, so the connections between tables become visible


jpraetorius:
Fixes #135

simonharrer:
A few comments:
- Hm, tests seem to be failing.
- Add in README and CHANGELOG
- Add a comment in the genereated DBML to say it was generated by the Data Contract CLI and its version. Maybe even add infos like the ID of the data contract.

simonharrer:
I'm not sure what types we should present there. I think it might be valuable using the logical types of the datacontract.yaml, but it could also be nice with the type mapping of a particular server technology (snowflake, postgres, etc.). Regardless, what types are used should be stated in the comments at the start of the generated file.

jpraetorius:
* it actually were `ruff` checks I failed to catch before :/ – ✅
* Adapted `CHANGELOG` and `README` ✅
* added some information that this is a generated file and who created it when for what datacontract ✅
* added the possibility to select a `--server` so we can convert the abstract types to concrete ones (this is currently reusing the existing SQL Export functionality, which is a bit of a wild import but seemed more reasonable than implementing this again) ✅

simonharrer:
Love it. Thanks for the contribution!

PR #216, opened 2024-05-22
Loosen OpenTelemetry dependency version constraint

janhicken:
When using the datacontract-cli as a library dependency, OpenTelemetry may well be already a dependency of the dependee. The previous dependency constraint allowed only other patch versions of OpenTelemetry to be used. However, other minor versions should also be compatible as of [OpenTelemetry's stability guarantees](https://github.com/open-telemetry/opentelemetry-python/blob/main/rationale.md#versioning-and-releasing).

As such, the dependency constraints for the OpenTelemetry OTLP exporters have been loosened to allow minor version upgrades.


simonharrer:
The tests are failing. There seems to be an issue with Soda and their open telemetry.

janhicken:
I did some investigation:

The deprecated `distutils` module has been removed in Python 3.12.

Soda is still using it in `soda.telemetry.soda_telemetry.py`:

```python
from distutils.util import strtobool
```

As it so happens, older versions of the OpenTelemetry SDK had a dependency to `setuptools`, which provides a drop-in replacement for `distutils`. As such, this issue never surfaced.

Newer versions of OpenTelemetry no longer have the `setuptools` dependency, so after the update, we're missing the `distutils` module. 

Actually, I think Soda should fix this issue upstream. However, I've added a dependency to `setuptools` in here as a workaround.

nigulable:
Is there smth that is blocking the merge? we are also waiting for this change. Thx

jochenchrist:
I opened a ticket for soda: https://github.com/sodadata/soda-core/issues/2091

jochenchrist:
Thank you for your contribution!

Issue #218, opened 2024-05-23
Export to jsonschema fails if object Field contains no subfields


hcrosse:
I'm encountering a `StopIteration` error [on this line](https://github.com/datacontract/datacontract-cli/blob/2908e2f999abd9ba49128a3d58b5f953c659ebf1/datacontract/export/jsonschema_converter.py#L57) when trying to export from a model that contains a Field of `type: object`  but contains no subfields. The jsonschema specification allows for empty objects so this seems like an unintentional error rather than deliberate decision.

Example model that fails
```
models:
  A:
    title: A
    description: ABC
    type: object
    fields:
       B:
        type: object
        description: BCD
```

simonharrer:
Fixed in #219

PR #219, opened 2024-05-23
Add jsonschema export handling for empty subfields

hcrosse:
Ref https://github.com/datacontract/datacontract-cli/issues/218

This PR allows for jsonschema export of object-typed fields that contain no subfields.


simonharrer:
Can you add a test, please? And perhaps add a CHANGELOG entry as well? That would be great. Otherwise, great catch. Thank you already. :-)

simonharrer:
Great work. Thank you again! :-)

Issue #122, opened 2024-03-28
datacontract import --format glue 


saugerDecathlon:
Hi, 

I am currently developing a function to init a contract from a Glue Table using this boto3 API. I would like to have your opinion about adding this feature to the CLI. 
It would look to something like

`datacontract import --format glue --database my_db --table my_table`

jochenchrist:
Interesting!

Supporting Data Catalogs (Dataplex, Glue Catalog, Purview, Unity Catalog, Hive, DataHub, Collibra) is definitely something on the roadmap.

Data contracts might become the source of metadata, and it will be natural to publish this information to technical data catalogs.

I am not quite sure, how important is import from catalogs vs. export/publish/push to catalogs and what will be the typical developer flows (contract-first vs. data-first). What do you think?

saugerDecathlon:
> Interesting!
> 
> Supporting Data Catalogs (Dataplex, Glue Catalog, Purview, Unity Catalog, Hive, DataHub, Collibra) is definitely something on the roadmap.
> 
> Data contracts might become the source of metadata, and it will be natural to publish this information to technical data catalogs.
> 
> I am not quite sure, how important is import from catalogs vs. export/publish/push to catalogs and what will be the typical developer flows (contract-first vs. data-first). What do you think?

Yes I agree for the export phase. However, in my view, numerous tables currently exist without a data contract, and introducing the import function would facilitate the implementation of the data contract on those 'legacy' tables.

simonharrer:
I second this @saugerDecathlon and I'd love to see a PR to add this import. :-)

jverhoeks:
I can help with this. I working on a project where the source is provided by lakeformation cross-account sharing to the etl accounts.  I already have some scrpt to extract the table config from glue for parquet and iceberg type tables.

This in combination with the export to dbt would speed up the pipeline development.

saugerDecathlon:
Great thank you @jverhoeks. This feature would be really helpful in my use cases

saugerDecathlon:
Did you had times to work on the feature @jverhoeks by any chance ? Otherwise I think I can push a PR on this topic by the end of the week

jverhoeks:
Hi, i created the code to extract the info but was checking how to implement this in datacontract-cli. 
This is simple code to extract the columns and hive partitions. 
Tested  on hive with parquet/json/csv  and iceberg tables. 

[...]

PR #230, opened 2024-05-29
Issues/122 support specifying tables for glue import

jpraetorius:
Allow to specify tables that should be imported from a given AWS Glue Database. 
Works the same way as we support it for BigQuery by the user giving a number of table names, but by default imports all table names.

This should solve the last discussion point in #122 

Also cleaned up the import documentation a bit.


jochenchrist:
LGTM

Issue #213, opened 2024-05-21
Library Usage Comes With Almost 1 GiB of Dependencies


janhicken:
When adding the `datacontract-cli` package as a dependency to a Python project, a lot of transitive dependencies get added. After adding the dependency, my application's Docker image grew from 330 MiB to 1.2 GiB in size.

My application only uses SodaCL in conjunction with a PostgreSQL database, however other frameworks like `pyspark` (340 MB), `pyarrow` (123 MB) and `deltalake` (75 MB) are integrated as well.

Would it be possible to split the packages per target technology like Soda does it? Instead, maybe [Extras](https://peps.python.org/pep-0508/#extras) can be used for this as well.

RobertLD:
@simonharrer I think it's likely worth splitting a lot of these packages into optional imports aka extras?

jochenchrist:
I think this is a fair point now, and we should add extras.

RobertLD:
> I think this is a fair point now, and we should add extras.

Drafting out the changes here #234

RobertLD:
@jochenchrist follow-up on this. Moving deltalake into an extra cut out 200MB of those larger deps I mentioned in the previous PR. I posted a PR here #240


I think 1.5Gb -> 300mb ought to be enough to close this issue out

jochenchrist:
I'd agree, with 1/5 of the dependency size, the library is much more efficient now :)

PR #234, opened 2024-05-31
#213 Split optional deps into extras to shrink dep tree

RobertLD:
Addressing the comment made in #213

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -25,12 +25,6 @@ dependencies = [
   "simple-ddl-parser==1.5.1",
   "soda-core-bigquery>=3.3.1,<3.4.0",
   "soda-core-duckdb>=3.3.1,<3.4.0",
-  "soda-core-sqlserver>=3.3.1,<3.4.0",
-  "soda-core-postgres>=3.3.1,<3.4.0",
-  "soda-core-snowflake>=3.3.1,<3.4.0",
-  "soda-core-spark[databricks]>=3.3.1,<3.4.0",
-  "databricks-sql-connector>=3.1.2,<3.2.0",
-  "soda-core-spark-df>=3.3.1,<3.4.0",
   # remove setuptools when https://github.com/sodadata/soda-core/issues/2091 is resolved
   "setuptools>=60",
   "snowflake-connector-python[pandas]>=3.6,<3.11",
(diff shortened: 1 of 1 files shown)


jochenchrist:
Thanks for your contribution, I agree, this is the way to go.

Futher todos:

- [ ] Not sure about `spark`. This should rather be bound to a server.
- [ ] Add bigquery?
- [ ] Add kafka? (-> requires spark...)
- [ ] Update README.md for an simple "all" installation, like `pip3 install datacontract-cli[bigquery,snowflake,databricks,spark,postgres]
- [ ] Make sure that basic funktionality (lint, diff, import, export?) works without any extra
- [ ] Create a CI/CD test matrix workflow for every extra

RobertLD:
> Thanks for your contribution, I agree, this is the way to go.
> 
> Futher todos:
> 
>     * [x]  Not sure about `spark`. This should rather be bound to a server.
> 
>     * [x]  Add bigquery?
> 
>     * [x]  Add kafka? (-> requires spark...)
> 
>     * [ ]  Update README.md for an simple "all" installation, like `pip3 install datacontract-cli[bigquery,snowflake,databricks,spark,postgres]
> 
>     * [ ]  Make sure that basic funktionality (lint, diff, import, export?) works without any extra
> 
>     * [ ]  Create a CI/CD test matrix workflow for every extra

In commit 02119adcc42fb0fd44e33ce6bd44ef9768c3aef1

jochenchrist:
I updated the order of the depdencies (reduired extras need to be defined first before being referenced)

RobertLD:
> I updated the order of the depdencies (reduired extras need to be defined first before being referenced)


* [x]  Not sure about `spark`. This should rather be bound to a server.

* [x]  Add bigquery?

* [x]  Add kafka? (-> requires spark...)

* [x]  Update README.md for an simple "all" installation, like `pip3 install datacontract-cli[bigquery,snowflake,databricks,spark,postgres]

* [x]  Make sure that basic funktionality (lint, diff, import, export?) works without any extra
** This was tested manually.
* [ ]  Create a CI/CD test matrix workflow for every extra
** Working on this one now

jochenchrist:
OK, I played around a bit with pytest markers, but it would significantly increase complexity with just a slight benefit. Let's drop this.

jochenchrist:
I updated the Dockerfile and added the Changelog

RobertLD:
I had a similar experience trying to configure the CI CD with pytest. The
two ideas I came up with are registering environmental variables that are
matched with pytest skip statements or having the tests be grouped by
folders that match the name of the extra and dynamically running pi test on
those depending on the extras that are installed

RobertLD:
> I updated the Dockerfile and added the Changelog I really appreciate the active contributors on this project 
I'll attach a dependency tree shortly to see what impact we exactly made by implementing the extras

jochenchrist:
OK, if you are happy with the reductions, I'd be happy to merge.

RobertLD:
> OK, if you are happy with the reductions, I'd be happy to merge.

The core installation now required 570M instead of 1.5Gb which is a pretty vast improvement. 

The largest of which are actual requirements for the core functionality. Perhaps finding a method of removing pyarrow from the core installation would be the final step. I'm happy with merging this as is; as it's still a 3x reduction

jochenchrist:
Happy to merge :)

Issue #105, opened 2024-03-21
datacontract import --format odcs


simonharrer:
Import from Open Data Contract Standard.

simonharrer:
Closing this as we now have a first draft.

PR #235, opened 2024-05-31
Import ODCS Contracts

jpraetorius:
- support basic information for `info`, `servers`, `terms` and `servicelevels`
- support for importing tables into models
- since there is no guarantee for the type that an ODCS Contract sets, we try to simply match what we can. For all other types people might specify, we use the ability of ODCS to have `customProperties` and allow people to specify specific mappings between 'their' types and the ones they want to have in the Datacontract by adding a magic `dc_mapping_typename` custom property.

Addresses #105 

@jochenchrist this is at least missing tests, but before heading on to them I'd just wanted to get your opinion, since I think for ODCS the mapping is not as straight forward as for some of the other specs I've seen. 
We might map more things depending on more magic strings for some of the SLA Properties for example, or we could try to coerce a columns examples to strings, but I'm not sure how far we want to take this, so input is welcome, before we make this a proper PR.

### DIFF
--- a/datacontract/cli.py
+++ b/datacontract/cli.py
@@ -201,6 +201,7 @@ class ImportFormat(str, Enum):
     glue = "glue"
     bigquery = "bigquery"
     jsonschema = "jsonschema"
+    odcs="odcs"
 
 
 @app.command(name="import")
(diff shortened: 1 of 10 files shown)


simonharrer:
Thanks! We'll start with this, and probably refine this along the way.

Issue #247, opened 2024-06-07
Improve spacing of 'Servers' section of data contract html page


Collinbrown95:
# Description

When I run `datacontract export datacontract.yaml --format html > datacontract.html` on the example provided in the repo, the Servers section of the html page places the headings and the values close together as shown in the screenshot below.

![Screenshot from 2024-06-06 23-36-51](https://github.com/datacontract/datacontract-cli/assets/8021046/bcbdcdcd-019f-40f9-a4af-ca6395c142e5)

# Proposed Solution

I wrapped the `dt` and `dd` elements in a `<div class="flex flex-col">`, and also added `px-3` (horizontal padding) to each div. The modified layout looks as follows.

![Screenshot from 2024-06-06 23-47-33](https://github.com/datacontract/datacontract-cli/assets/8021046/9460915b-aa0a-4ce1-9ece-114d4888b625)

I'll open a PR with the implementation described above, happy to revise this approach if one of the maintainers has another idea.

jochenchrist:
Thanks for reporting and the PR.

I also fixed the underlying root cause:
The partial was missing in the tailwindcss processor (we need to think about a better solution here), as the gap-x-4 was not picked up.

https://github.com/datacontract/datacontract-cli/commit/bc8bec3e0858802d86dced316c8182d3a34d09ff

PR #248, opened 2024-06-07
feat(server.html): Small CSS changes to improve spacing and layout.

Collinbrown95:
# Description

Addresses https://github.com/datacontract/datacontract-cli/issues/247

- Add div with class 'flex flex-col' to each component.
- Introduce horizontal padding to each component with `px-3` class.


jochenchrist:
Thanks for your contribution.

PR #277, opened 2024-06-21
Add Spark SructType exporter

pierre-monnet:
Add the ability to export Spark StructType.

- [x] Tests pass
- [x] README.md updated


simonharrer:
I forgot to update the CHANGELOG. Can you rebase your changelog entry so that it appears in the unreleased section again, please?

README.md
```diff
@@ -765,6 +766,9 @@ The export function converts the logical data types of the datacontract into the
 if a server is selected via the `--server` option (based on the `type` of that server). If no server is selected, the
 logical data types are exported.
 
+#### Spark
+
+The export function converts the data contract specification into a StructType Spark schema.
```
simonharrer, line 771: Can you add a link to the spark schema in the docs? So people can go to the according docs for that StructType?

simonharrer:
Thanks for the PR. Looks nice. My question: should the output not be model dependent? So one would have to call the export per model, to get a structtype for that model? Or is it typical to put multiple struct types into a parent dict? I'm raising this as I'm thinking regarding the integration of our CLI tool with other tools, and with this, one would always has to extract a specific structtype again for further processing.

README.md
pierre-monnet: I added some details about Spark Data Types

pierre-monnet:
> Thanks for the PR. Looks nice. My question: should the output not be model dependent? So one would have to call the export per model, to get a structtype for that model? Or is it typical to put multiple struct types into a parent dict? I'm raising this as I'm thinking regarding the integration of our CLI tool with other tools, and with this, one would always has to extract a specific structtype again for further processing.

I changed the SparkExporter to return a string "picture" of python code generated.
I keep the function that returns a dict for programmatic usage

datacontract/cli.py
```diff
@@ -211,6 +211,7 @@ class ImportFormat(str, Enum):
     jsonschema = "jsonschema"
     odcs = "odcs"
     unity = "unity"
+    spark = "spark"
```
jochenchrist, line 214: This importer format seems wrong here (and causes a merge conflict)

pierre-monnet: I resolved conflict.
By the way it was a mistake to put spark in `ImportFormat`

README.md
```diff
@@ -765,6 +766,10 @@ The export function converts the logical data types of the datacontract into the
 if a server is selected via the `--server` option (based on the `type` of that server). If no server is selected, the
 logical data types are exported.
 
+#### Spark
+
+The export function converts the data contract specification into a StructType Spark schema.  
+Spark DataFrame schema is defined as StructType. For more details about Spark Data Types please see [the spark documentation](https://cloud.google.com/docs/authentication/provide-credentials-adc#how-to)
```
simonharrer, line 772: The link seems wrong

pierre-monnet: Fixed 😶‍🌫️

simonharrer:
Fine for me. We should, however, document, that we basically generate python code here that uses pyspark.

pierre-monnet:
> Fine for me. We should, however, document, that we basically generate python code here that uses pyspark.

I added this detail in the README

simonharrer:
Thanks! :-)

PR #287, opened 2024-06-24
doc: add a custom exporter example

teoria:
- [x] Tests pass
- [x] README.md updated


jochenchrist:
@simonharrer should we move this to another documentation subpage?

simonharrer:
Not now. I would do this when we rework the whole readme and navi.

PR #293, opened 2024-06-25
fix: load exporters dynamically

teoria:
this PR implements dynamic module load to exporter to avoid errors when the user install a specific dependence like

`datacontract-cli[databricks]==0.10.8`

- [x] Tests pass

### DIFF
--- a/datacontract/export/exporter_factory.py
+++ b/datacontract/export/exporter_factory.py
@@ -1,21 +1,6 @@
-from datacontract.export.avro_idl_converter import AvroIdlExporter
-from datacontract.export.bigquery_converter import BigQueryExporter
-from datacontract.export.dbml_converter import DbmlExporter
-from datacontract.export.dbt_converter import DbtExporter, DbtSourceExporter, DbtStageExporter
-from datacontract.export.avro_converter import AvroExporter
+import importlib
+import sys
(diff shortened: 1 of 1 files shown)


jochenchrist:
Is this ready to merge?

I understand this is necessary, but the code base becomes pretty complex :/

simonharrer:
I feel that going back to a simple if/else might be much simpler as we have to create an if per exporter anyway. We can still use the common interface. Thoughts?

teoria:
@jochenchrist 

> Is this ready to merge?
> 
> I understand this is necessary, but the code base becomes pretty complex :/

Ready to merge.

I don't know how to cover by test this issue( the user don't install the 'avro' and run a AvroImporter)

The dynamic dependency loader add a complexity in this module but enabled to create a external config file in the future.

Eg.
Config.yaml with the list of supported importers and exporters.

New features (importer or exporters) can be add without any changes in previous code and custom importers and exporters can be add like I wrote in the doc.

teoria:
@simonharrer 
> I feel that going back to a simple if/else might be much simpler. We can still use the common interface. Thoughts?

I understand your feeling 
Your are correct If/else is simpler
But the main ideia of this refact is make the growth easier 
It's not about creation, it's about maintainability.

PR #309, opened 2024-07-03
Update duckdb.py for s3 connection by utilizing session_token for short term credentials

masoodkhalid32:
S3 connection would not work with only access_key_id and secret_access_key if it short_term_credentials. adding session _token in the connection setup for s3 with short term credentials.

- [ ] Tests pass
- [ ] ruff format


jochenchrist:
Thanks for your contribution.

Issue #104, opened 2024-03-21
datacontract import --format dbt


simonharrer:
Out of #103 came the idea of having an import of dbt models to a datacontract.yaml

```
datacontract import --format dbt models.yaml
```

emirkmo:
I already do something like this for import, creating a datacontract.yaml given a dbt project, but was using the "schema" field instead of the "model" field, with a custom schema type. (Slightly off-topic, but `schema` was much more widely understandable than `models` in our workshops. Just some feedback I can provide on the specification issue).

However our code is/was quite specific to the format of the dbt projects we allowed. To do it properly, one would want to parse & use the manifest.json file from a dbt project. It is the most straightforward way of working with dbt projects generically.

You would go into dbt Nodes in the manifest, and for every `resource_type` of `model` import the columns, data_types if given, descriptions if given, etc. The only difficulty is mapping the data_types to the supported ones in datacontract spec. Hence why physical model specific `schema` might make more sense for the import.. As a first step though, the `model` in `models` could just not provide the `data_type` or provide the dbt one if it matches. 

(For parsing the manifest, Dagster-dbt does this as well, and the code is Apache-2 Licensed, if you are looking for inspiration).
The import is something I can contribute on, if the implementation sounds ok.

pixie79:
I think the later is fine. As I presume most people with more than a few dbt models split them into a model per file otherwise it gets quite unwieldily very quickly.
Either that or parse them all but allow an input to specify which models you want to include in the data contract as it could be you want to or three for a specific contract?

emirkmo:
> I think the later is fine. As I presume most people with more than a few dbt models split them into a model per file otherwise it gets quite unwieldily very quickly. 

This does not match my experience with larger dbt projects. But one or several models can logically co exist and be part a data contract so it is fine anyway?

torbenkeller:
I'm looking into this right now

simonharrer:
Awesome! I assigned you the issue. :-)

teoria:
i've been working with dbt, maybe I can help

torbenkeller:
@jochenchrist Was working on other things the last weeks, sorry. But I will continue on this.

@teoria sounds good, if you want we can pair program to get this ready

torbenkeller:
@teoria you can contact me on the datacontract slack channel

teoria:
Nice!

teoria:
first version of dbt manifest importer
[...]

PR #317, opened 2024-07-09
feat: add dbt manifest importer

teoria:
import dbt manifest file
issue: https://github.com/datacontract/datacontract-cli/issues/104
 
- [x] Tests pass
- [x] ruff format 


datacontract/imports/dbt_importer.py
```diff
@@ -0,0 +1,116 @@
...
+    description: Optional[str] = None
+    data_type: Optional[str] = None
+    meta: Optional[dict] = None
+    constraints: Optional[str] = None
+    quote: Optional[str] = None
+    tags: Optional[str] = None
+
+    def __init__(self, node_column) -> None:
+        self.name = node_column.get("name", "")
+        self.description = node_column.get("description", "")
```
jochenchrist, line 83: Any specific reason, why default is "" instead of None?

datacontract/imports/dbt_importer.py
```diff
@@ -0,0 +1,116 @@
...
+
+class DbtModel:
+    name: str
+    database: str
+    schema: str
+    description: Optional[str] = None
+    unique_id: str
+    tags: Optional[str] = None
+
+    def __init__(self, node) -> None:
```
jochenchrist, line 100: What type is node?

datacontract/imports/dbt_importer.py
```diff
@@ -0,0 +1,116 @@
...
+        self.name = node.get("name")
+        self.database = node.get("database")
+        self.schema = node.get("schema")
+        self.description = node.get("description")
+        self.display_name = node.get("display_name")
+        self.unique_id = node.get("unique_id")
+        self.columns = []
+        self.tags = node.get("tags")
+        self.add_columns(node.get("columns").values())
```
jochenchrist, line 109: Can node.get("columns") be None?

datacontract/imports/dbt_importer.py
teoria: if None the render will remove from output yaml

datacontract/imports/dbt_importer.py
teoria: dict

datacontract/imports/dbt_importer.py
jochenchrist: Any reason why you don't want to remove the description then?

datacontract/imports/dbt_importer.py
jochenchrist: then let's add the type?

datacontract/imports/dbt_importer.py
teoria: i removed de default value ""

jochenchrist:
Thanks for your contribution!

Issue #312, opened 2024-07-05
Pyspark dependency is required despite marked as optional


janhicken:
The `pyspark` dependency has been changed to an optional dependency, installed using either the `databricks` or `kafka` extra.

However, the [`data_contract` module](/datacontract/datacontract-cli/blob/v0.10.9/datacontract/data_contract.py#L7) imports `pyspark` explicitly. This will fail when `pyspark` is not installed:

```
data_contract.py:7: in <module>
    from pyspark.sql import SparkSession
E   ModuleNotFoundError: No module named 'pyspark'
```

As the `SparkSession` imported is used for type-checking only, maybe this can be fixed easily.

Furthermore, the same issue can be found in the [`check_soda_execute` module](/datacontract/datacontract-cli/blob/v0.10.9/datacontract/engines/soda/check_soda_execute.py#L3).

RobertLD:
https://docs.python.org/3/library/typing.html#typing.TYPE_CHECKING

This looks to be the accepted solution to problems like this; but it's not very elegant. I can put a fix up within a few days but if someone else wants to go ahead; should only be a couple minutes.

RobertLD:
Opened a pull request for this using the suggested solution via the Typing module

jochenchrist:
PR merged.

RobertLD:
> PR merged.

That PR had an issue that caused CI/CD to fail. Please see the follow up 
https://github.com/datacontract/datacontract-cli/pull/320


Sorry about that again

PR #319, opened 2024-07-09
Only import pyspark when actually required, otherwise use it for LSP

RobertLD:
Resolves issue reported in #312 

- [x] Tests pass
- [x] ruff format


jochenchrist:
Thanks for your contribution.

RobertLD:
@jochenchrist  Hey this only passed my local tests and not the CI/CD one. I'm still investigating the difference at the moment

PR #320, opened 2024-07-09
Fix/312 only import pyspark if type checking

RobertLD:
- [x] Tests pass


RobertLD:
@jochenchrist This PR addresses the test failures as a result of using TYPE_CHECKING

PR #326, opened 2024-07-15
Import_spark

pierre-monnet:
- [x] Tests pass
- [x] ruff format


jochenchrist:
LGTM

tests/test_import_spark.py
```diff
@@ -0,0 +1,162 @@
...
+
+expected = """
+dataContractSpecification: 0.9.3
+id: my-data-contract-id
+info:
+  title: My Data Contract
+  version: 0.0.1
+servers:
+  local:
+    type: dataframe
```
simonharrer, line 21: I think we should not add this here, as there is no server type `dataframe` at the moment at datacontract.com

jochenchrist: Yes, there is. At least in the schema.

PR #335, opened 2024-07-17
Fix glue_table parameter not filtering tables

roykoand:
- [ ] Tests pass
- [x] ruff format


jochenchrist:
Thank you for spotting this bug and fixing

PR #339, opened 2024-07-21
feat: Propagate all env vars with DATACONTRACT_SNOWFLAKE_ prefix

tarys:
[Soda documentation](https://docs.soda.io/soda/connect-snowflake.html) states that [all parameters](https://docs.snowflake.com/developer-guide/python-connector/python-connector-api#functions) from the Snowflake Python Connector are supported.
Thus, we only need to propagate them from environmental variables to the underlying Soda configuration.

This change will allow a lot of flexibility when connecting to Snowflake.
For example, OAuth and RSA key authentication, custom session timezone, etc.

### DIFF
--- a/README.md
+++ b/README.md
@@ -588,14 +588,31 @@ models:
 ```
 
 #### Environment Variables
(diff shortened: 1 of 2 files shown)


jochenchrist:
Great contribution, very elegant. Thanks.

tarys:
Thanks a lot!
Tried to make it both backward compatible and future proof.

PR #347, opened 2024-07-25
Local support delta

### LINKED ISSUE #338: Delta table support for server type local is not implemented yet, it only checks for azure  and then directly goes to AWS s3
I think the control flow does this in line number 52 of https://github.com/datacontract/datacontract-cli/blob/main/datacontract%2Fengines%2Fsoda%2Fconnections%2Fduckdb.py

Happycipher:
- [x] Tests pass
- [x] ruff format
Description- Added an extra condition to support delta tables in server type "local" and placed the default server type "s3" under an if condition. Added test cases for the same.
Fixes #338


simonharrer:
Thanks for you contribution!

PR #348, opened 2024-07-26
Improve dbml export

dstachowicz:
- Recognise 1-to-1 relationships and export them as such in DBML
- Fix parse errors in a common DBML renderer (even though produced markup was valid)
- Minor improvement wrt. to quotation marks


jochenchrist:
LGTM

Issue #346, opened 2024-07-24
Breaking method  is not working with bigint columns


teoria:
When you compare 2 contracts with bigint and long column must be equal.

```

datacontract breaking  contract_v1.yaml contract_v2_glue.yml

1 breaking changes: 1 error, 0 warning
error   [field_type_updated] at contract_v2_glue.yml
        in my_table.fields.my_column.type
            changed from `bigint` to `long`
```

roykoand:
@teoria I don't think that AWS Glue supports `long` data type. I've only seen `bigint` data type in their UI once tried to create a table:
<img width="631" alt="image" src="https://github.com/user-attachments/assets/9c070ccb-5159-4954-9806-8f5040ecf1a5">

<img width="657" alt="image" src="https://github.com/user-attachments/assets/41c8690c-3b14-4c37-8db3-cfab57a1fb69">

teoria:
when i use the glue importer the code cast bigint to long
then when i call breaking  the cli identify the type update.
the breaking method need ignore bigint->long or the import need keep bigint type

roykoand:
@teoria yes, I agree. Data contract should keep `bigint` data type. This line is a culprit:
https://github.com/datacontract/datacontract-cli/blob/56f4dbe4469c65c03b42fc6464258b6e6dd1e193/datacontract/imports/glue_importer.py#L268

teoria:
or breaking ignore the bigint->long
@jochenchrist what do you think ?

jochenchrist:
lets change the importer to keep bigint

jochenchrist:
fixed with #351.

jochenchrist:
Will be part of release v0.10.12

PR #351, opened 2024-07-27
AWS Glue import refactoring

roykoand:
- [x] Tests pass
- [x] ruff format


datacontract/imports/glue_importer.py
```diff
@@ -109,18 +109,17 @@ def get_glue_table_schema(database_name: str, table_name: str):
                     "Name": pk["Name"],
                     "Type": pk["Type"],
                     "Hive": True,
-                    "Comment": "Partition Key",
+                    "Comment": pk.get("Comment"),
```
roykoand, line 112: `"Comment": "Partition Key"` is overriding the column comment, I suggest to remove it and leave `required: true` as the only signal of the partition column

datacontract/imports/glue_importer.py
```diff
@@ -253,30 +267,20 @@ def map_type_from_sql(sql_type: str) -> str:
...
-    if sql_type.startswith("date"):
-        return "date"
-    if sql_type.startswith("decimal"):
-        return "decimal"
-    return "variant"
+
+    type_mapping = {
+        "string": "string",
+        "int": "int",
+        "bigint": "bigint",
```
roykoand, line 274: Issue https://github.com/datacontract/datacontract-cli/issues/346 with `bigint` -> `long` conversion has been fixed. Also, some data types that are not used by Glue have been removed - `short`, `byte`, `text`. 
Also, `integer` has been renamed to `int` to be consistent with Glue (not backward compatible change).
Notice that `variant` was renamed to `unknown`, this is what Glue uses when something went wrong
image

datacontract/imports/glue_importer.py
```diff
@@ -140,8 +139,13 @@ def import_glue(
     if table_names is None:
         table_names = get_glue_tables(source)
 
+    server_kwargs = {"type": "glue", "account": catalogid, "database": source}
+
+    if location_uri:
+        server_kwargs["location"] = location_uri
```
roykoand, line 145: `location_uri` with `"None"` string is considered to be None value while desearalizing to YAML.
Therefore Glue database without a location creates invalid data contract:
```
datacontract-cli git:(56f4dbe) ✗ datacontract test test1.yaml 
Testing test1.yaml
WARNING:root:Data Contract YAML is invalid. Validation error: data.servers.production must be valid exactly by one definition (0 matches found)
ERROR:root:Run operation failed: [lint] Check that data contract YAML is valid - None - failed - data.servers.production must be valid exactly by one definition (0 matches found) - datacontract
╭────────┬────────────────────────────────────────┬───────┬────────────────────────────────────────────────────╮
│ Result │ Check                                  │ Field │ Details                                            │
├────────┼────────────────────────────────────────┼───────┼────────────────────────────────────────────────────┤
│ failed │ Check that data contract YAML is valid │       │ data.servers.production must be valid exactly by   │
│        │                                        │       │ one definition (0 matches found)                   │
╰────────┴────────────────────────────────────────┴───────┴────────────────────────────────────────────────────╯
1) data.servers.production must be valid exactly by one definition (0 matches found)
```
Let's ignore destination param if it's not specified

jochenchrist: Is the location optional in Glue?

roykoand: @jochenchrist wow, thanks for the fast feedback! Yes, it's optional.
That's the location description you get in their UI:
image
and confirming I do have some Glue databases without a location specified

datacontract/imports/glue_importer.py
```diff
@@ -186,27 +184,43 @@ def create_typed_field(dtype: str) -> Field:
     """
     field = Field()
```
roykoand, line 185: Lots of Glue data types are not yet implemented:
- `binary`
- `char`
- `interval`
- `smallint`
- `tinyint`
- `union`
etc.

I am not quite sure how to add them cause the supported list of data types:
```
['number', 'decimal', 'numeric', 'int', 'integer', 'long', 'bigint', 'float', 'double', 'string', 'text', 'varchar', 'boolean', 'timestamp', 'timestamp_tz', 'timestamp_ntz', 'date', 'array', 'map', 'object', 
'record', 'struct', 'bytes', 'null']
```
https://datacontract.com/#data-types

Can I just extend this list of supported data types with the ones above? Thanks!

jochenchrist: Please try to map Glue Types to a logical data type (https://datacontract.com/#data-types
) that makes most sense. 

You can use a config object to specify the physical type:

```
type: bytes
config:
  glueType: binary
```

Some types (esp. union) may not be supported, add a warning here.

datacontract/imports/glue_importer.py
jochenchrist: OK, do you still get this issue? `location` was already optional in the schema.
If you still have this error, can you post the server object that you get.

roykoand: @jochenchrist yup, it's optional in Pydantic model. However, the JSON schema of specification does not allow anything except URI string: https://github.com/datacontract/datacontract-specification/blob/816e6ae21e7ebdd2b9261734bd1a7ee47388574c/datacontract.schema.json#L261
Therefore, Glue database without location will create an invalid contract with `location: None`.
If you validate this contract with `datacontract test` you will get the error above

roykoand: @jochenchrist yup, it's optional in Pydantic model. However, the JSON schema of specification does not allow anything except URI string: https://github.com/datacontract/datacontract-specification/blob/816e6ae21e7ebdd2b9261734bd1a7ee47388574c/datacontract.schema.json#L446
Therefore, Glue database without location will create an invalid contract with `location: None`.
If you validate this contract with `datacontract test` you will get the error above.
We can make the

datacontract/imports/glue_importer.py
roykoand: Amazing! Thank you, I will keep that in mind for the next PRs

datacontract/imports/glue_importer.py
jochenchrist: If the location is not set, then the data contract YAML should be valid (location is not a required property):
```
servers:
  production:
    type: glue
    account: '123456789012'
    database: test_database
```

roykoand: @roykoand Yes, sorry for the confusion. It's fixed by this PR but was not working before. Thanks!

datacontract/imports/glue_importer.py
jochenchrist: ok

roykoand: @jochenchrist Thanks for the review! Is there anything else that needs to be done in this PR?

jochenchrist:
Thanks for your contribution. Happy to merge.

PR #352, opened 2024-07-27
JSONSchema importer refactor for easier maintenance

johannesrave:
- REMOVED behaviour where types would be set as optional/not required if the type-schema was an array containing a `null` entry. this doesn't appear to be part of the JSONSchema spec, but we could be lenient about it and keep it?
- KEEPS the behaviour to select the first type in a list of types, so merges union-types into the first non-null-type. we should probably raise an error instead like we do with union types in arrays!
- UNWRAPPED raised exceptions for easier debugging.  if the wrapping was intentional i'll put it back in.
- unified handling of Fields and Definitions in the JSONSchema importer
- untangled logic for type-detection and required-detection
- reduced indentation where adequate

- [+] Tests pass
- [+] ruff format
- [-] CHANGELOG.md entry added (irrelevant, no features added)

### DIFF
--- a/datacontract/imports/jsonschema_importer.py
+++ b/datacontract/imports/jsonschema_importer.py
@@ -18,6 +18,29 @@ def import_jsonschema(data_contract_specification: DataContractSpecification, so
     if data_contract_specification.models is None:
         data_contract_specification.models = {}
 
+    json_schema = load_and_validate_json_schema(source)
+
+    title = json_schema.get("title", "default_model")
+    description = json_schema.get("description")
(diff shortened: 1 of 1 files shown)


simonharrer:
> KEEPS the behaviour to select the first type in a list of types, so merges union-types into the first non-null-type. we should probably raise an error instead like we do with union types in arrays!

Does JSON Schema really support this? I thought one would need to use `oneOf` or `anyOf` as a wrapping element. I think the array for `items` can never happen.

simonharrer:
LGTM.

PR #366, opened 2024-08-07
Redesign dbt importer

AlexanderHansson4225:
- [x] Tests pass
- [x] ruff format

What changes were made:

We aimed to make the dbt importer more robust as well as customizable, throught the following.

- Use `dbt-core` contract Node definitions (which are dataclasses, similar to pydantic models) instead of inventing our own classes to maintain
- Implement import args as a dictionary for customizing the import from a manifest.json, without having to re-write an `Importer` from scratch for parsing your own dbt manifest.
    - For example: In addition to importing only certain named models, one might want to import seeds, or import models with a specific tag, etc. The way it was hard-coded made it so that one had to just create their own `DbtImporter` from scratch for customization even when staying within the Data Contract Spec.

- Fix/add correct type hints
 -Add docstrings for info

- Minor refactor of `import_dbt_manifest` and `create_fields` (list comprehension, flag for dbt resource type, some comments, etc.)

Why we made these changes:

To allow for more customization of the dbt importer. We purposefully kept it pretty close to the original as far as implementation goes, however this version is more open to further development. In our case, the existing `DbtColumn` and `DbtModel` classes were missing a lot of the critical fields we needed from the dbt manifest file, which meant that we had to write an importer like in the current PR anyway.

If this approach seems good, we would be happy to further generalize the dbt importer to make it easier to either Wrap with a custom importer, or use as is with some built in customizability.

In our own custom DBT importer, we have additional `import_args` called "transformations" which are allowed to be callables that take a `DataContractSpecification` and a dbt Manifest and return a `DataContractSpecification`. This means that we can parse metadata from the Manifest, and automatically fill in more fields in the data contract--using just the Dbt manifest. (Think about everything from contact information to Quality or freshness generated directly from the dbt manifest file).

However, since this sort of parsing is very custom, we leave the base dbt importer as a generic implementation that allows for custom transformations, which is much easier to implement as a strategy pattern instead of re-writing the importer from scratch for each time.

Who we are:

We are Data Engineers at Axis Communications AB.


jochenchrist:
@teoria You created the original dbt importer. Could you review this PR?

jochenchrist:
@AlexanderHansson4225 can you have a look ath the merge conflict (coming from another PR)

AlexanderHansson4225:
> @AlexanderHansson4225 can you have a look ath the merge conflict (coming from another PR)

@jochenchrist  The merge conflicts have been resolved

AlexanderHansson4225:
As I am a student worker, on my last day today, my supervisor @emirkmo will take over this pull request from now on. He has push access to my fork.

teoria:
@AlexanderHansson4225 thanks for the contribution.
the code looks better now. The only thing I'm not sure about is whether it's worth creating a dependency on the dbt package just to reuse the Manifest parser or would it be better to bring the parser to the project.

what do you think @jochenchrist  ?

simonharrer:
I like it. dbt support is important. If the dependecy makes trouble, we can throw it out later.

Issue #379, opened 2024-08-16
dbml import


simonharrer:
Import from a dbml file

The CLI already supports dbml export

simonharrer:
We could use https://github.com/Vanderhoof/PyDBML

PR #389, opened 2024-08-21
Add support for importing from DBML

jpraetorius:
Addresses #379 

- ingest basic project data
- import tables and columns as best we can
- allow filtering for tables and schemas
- still TODO: References

- [X] Tests pass
- [X] ruff format


jochenchrist:
LGTM

Issue #256, opened 2024-06-10
datacontract export --format Avro - use 'required' attribute


gpthome:
Request - use the "required" field in the model to generate AVRO output accordingly.  I believe when set to _false_, use UNION with NULL being first.

If model is defined like this...

```
      tags:
        type: array
        items: 
          type: string
        default: []
        required: false
      removal_dt:
        type: timestamp
        required: false
      barcode:
        type: string
        required: false
```

Generate output similar to this...

```
        {
            "name": "tags",
            "type": [
                "null",
                {
                    "type": "array",
                    "items": "string"
                }
            ],
            "default": null
        },
        {
            "name": "removal_dt",
            "type": [
                "null",
                {
                    "type": "long",
                    "logicalType": "timestamp-millis"
                }
            ]
        },
        {
            "name": "barcode",
            "type": [
                "null",
                {
                    "type": "string",
                }
            ]
        },
```

Please include other data types like **int** as well.  For the array, I'm not 100% sure if the generated AVRO output "default" should be null or "[ ]"

gpthome:
jsamson1's issue/PR above addresses the concerns of this issue.  I'm closing this request.  Thanks!

PR #390, opened 2024-08-22
Issues/256 Support Required Fields for Avro Schema Export

jsamson1:
- [X] Tests pass
- [X] ruff format

Resolves issue #256

This PR adds support for required fields when exporting to Avro. If not specified, the required attribute defaults to true to preserve prior behavior. Added new test function to test_export_avro.py and necessary test schemas (datacontract_test_required.avsc & datacontract_test_required.yaml)


datacontract/export/avro_converter.py
```diff
@@ -40,7 +40,10 @@ def to_avro_field(field, field_name):
     avro_field = {"name": field_name}
     if field.description is not None:
         avro_field["doc"] = field.description
-    avro_field["type"] = to_avro_type(field, field_name)
+    if field.required is None:
+        field.required = True
```
jochenchrist, line 44: I don't get this. Why do you set field.required to True if the property is set (cloud also be false, couldn't it)?

jsamson1: This part of the code was to handle cases where the yaml field did not have a required attribute set (so neither true or false). My initial thought was to set that field to true in those cases since all fields in avro are required (unless specified otherwise). I was able to remove this part and still handle those cases correctly. The test I added shows the three different cases

jochenchrist (approved):
LGTM

PR #397, opened 2024-08-27
Feature/avro export enum support

cners-follett:
- [X] Tests pass
- [X] ruff format

When importing an avro file, we noticed the name attribute from the enum (previously implemented) wasn't being respected.  Small code adjustment for that, as well as setting test expectations accordingly.

Also added support of enum on export.
Previously, if you had an enum declared, upon export, it would look like this: 
image

After:
yaml: 
image
avsc:
image


tests/test_import_avro.py
```diff
@@ -78,6 +78,7 @@ def test_import_avro_schema():
         type: string
         required: true
         description: order status
+        name: Status
```
jochenchrist, line 81: `name` is not a valid property in the https://datacontract.com/#field-object Field Object.

Do you mean `title`?

cners-follett: Thank you so much for pointing out the Field Object & associated properties of it. I will make that change - I also saw I broke a test, so will review that as well before updating the PR!

jochenchrist: This part has not been resolved, has it?

cners-follett: You're right - I fixed up export but not import.  I made the test fail and then fixed the code. Keen eye. Thank you!

jochenchrist:
LGTM, Thanks for your contribution!

Issue #398, opened 2024-08-29
Add support for SQLAlchemy as export format


characat0:
Thanks for the great tool
It would be great to have support for [SQLAlchemy](https://www.sqlalchemy.org) as export format, since it is a widespread tool and ORM.
It also integrates well with [alembic](https://alembic.sqlalchemy.org/en/latest/tutorial.html), which allows us to autogenerate data migrations.

PR #399, opened 2024-08-30
Add support for SQLAlchemy as export format

characat0:
Closes https://github.com/datacontract/datacontract-cli/issues/398

# Summary
In this PR I propose to add basic support for SQLAlchemy, we can discuss further details here.

My particular use case is focused on databricks sql connection, but I would be happy to receive feedback from people with other use cases

## Description

Table names and comments are supported, as well as field comments. Table schema support is still limited to schema only, with support for fully qualified schemas (`catalog.schema`) on autogenerated migrations depending on https://github.com/sqlalchemy/alembic/issues/1526.

### Data Types
I've currently implemented support for the following primitive data types:
* string: String with maxLength
* text: String with maxLength
* varchar: VARCHAR with maxLength
* number, decimal, numeric: Numeric with precision and scale
* int, integer: Integer
* long, biting: BigInteger
* float: Float
* double: Double
* boolean: Boolean
* timestamp: TIMESTAMP
* timestamp_tz: TIMESTAMP with tz=True
* timestamp_ntz: TIMESTAMP_NTZ (only supported in databricks)
* date: Date
* bytes: LargeBinary with maxLength

I've also added support for arrays of these type of primitives.

- [x] Tests pass
- [x] ruff format


jochenchrist:
Thanks for this PR. I think, this is a very valuable contribution.

Please add a test case for this exporter.

characat0:
@jochenchrist thanks for the feedback, I've added some test cases inspired by the Pydantic export implementation

jochenchrist:
can you please also update the README and the CHANGELOG?

PR #422, opened 2024-09-16
Update BQ map method

ffernandez92:
- [x] Tests pass
- [x] ruff format


datacontract/export/bigquery_converter.py
```diff
@@ -79,44 +79,53 @@ def to_field(field_name: str, field: Field) -> dict:
...
-    elif type_str == "null":
+    elif field_type.lower() == "null":
         logger.info(
-            f"Can't properly map {field_name} to bigquery Schema, as 'null' is not supported as a type. Mapping it to STRING."
+            f"Can't properly map {field.title} to bigquery Schema, as 'null' \
+                 is not supported as a type. Mapping it to STRING."
         )
         return "STRING"
     else:
         raise DataContractException(
```
simonharrer, line 125: I do not think we need this anymore. We could simply import, put type "custom" in and add the imported type under config.bigqueryType.

tests/fixtures/bigquery/export/datacontract.yaml
```diff
@@ -180,4 +180,10 @@ models:
             Field2:
               type: double
               required: true
-              description: a double field
\ No newline at end of file
+              description: a double field
+      custom_type_field:
+        type: string
```
simonharrer, line 185: I'd rather map to custom and use the config object instead.

datacontract/export/bigquery_converter.py
ffernandez92: I think if there is a `config.bigqueryType` the exception won't trigger either way because of the first `if` statement of the method right?

simonharrer:
Love it. Can you add a changelog? Then, I'll merge it.

jochenchrist:
I'll add a changelog

Issue #436, opened 2024-09-20
[Bug]: Support schema-validation checks with custom `precision` and `scale` for `DecimalType` with `dataframe` server


FredrikBakken:
Schema validation with `dataframe` server currently fails when schema uses custom `precision` and `scale` values for `DecimalType`.

jochenchrist:
PR was merged.

PR #437, opened 2024-09-20
feat: add sql type converter for `dataframe` server

FredrikBakken:
- [ ] Tests pass
- [X] ruff format

Issue reference: https://github.com/datacontract/datacontract-cli/issues/436

### DIFF
--- a/datacontract/export/sql_type_converter.py
+++ b/datacontract/export/sql_type_converter.py
@@ -113,6 +113,8 @@ def convert_type_to_postgres(field: Field) -> None | str:
 # dataframe data types:
 # https://spark.apache.org/docs/latest/sql-ref-datatypes.html
 def convert_to_dataframe(field: Field) -> None | str:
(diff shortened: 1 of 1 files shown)


datacontract/export/sql_type_converter.py
```diff
@@ -108,6 +110,44 @@ def convert_type_to_postgres(field: Field) -> None | str:
     return None
 
 
+# dataframe data types:
+# https://spark.apache.org/docs/latest/sql-ref-datatypes.html
+def convert_to_dataframe(field: Field) -> None | str:
+    type = field.type
```
jochenchrist, line 116: I think we should also support a config option `config.dataframeType`:

Could you add this?

    if field.config and "dataframeType" in field.config:
        return field.config["dataframeType"]

jochenchrist:
LGTM. Thanks for your contribution.

PR #419, opened 2024-09-11
fix: add support for `precision` and `scale` for `DecimalType`

FredrikBakken:
- [ ] Tests pass
- [X] ruff format
- [X] README.md updated (if relevant)
- [ ] CHANGELOG.md entry added

Add `precision` and `scale` values to Spark `DecimalType`.

PR #450, opened 2024-10-06
fix: default `precision` and `scale` parameters for `DecimalType`

FredrikBakken:
- [ ] Tests pass
- [X] ruff format

In a previous PR (https://github.com/datacontract/datacontract-cli/pull/419), I did not consider the case where `precision` and `scale` are not provided in the data contract. This change allows defining `type: decimal` without adding `precision` and/or `scale` values.

### DIFF
--- a/datacontract/export/spark_converter.py
+++ b/datacontract/export/spark_converter.py
@@ -128,7 +128,7 @@ def to_data_type(field: Field) -> types.DataType:
     if field_type in ["string", "varchar", "text"]:
         return types.StringType()
     if field_type in ["number", "decimal", "numeric"]:
-        precision = field.precision if field.precision is not None else 10
+        precision = field.precision if field.precision is not None else 38
         scale = field.scale if field.scale is not None else 0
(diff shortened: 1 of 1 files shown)


datacontract/export/spark_converter.py
```diff
@@ -128,7 +128,9 @@ def to_data_type(field: Field) -> types.DataType:
     if field_type in ["string", "varchar", "text"]:
         return types.StringType()
     if field_type in ["number", "decimal", "numeric"]:
-        return types.DecimalType(precision=field.precision, scale=field.scale)
+        precision = field.precision if field.precision is not None else 10
```
jochenchrist, line 131: default is 38.

Reference: https://datacontract.com/#field-object

jochenchrist:
Thanks for your contribution. Have a look at the comment.

datacontract/export/spark_converter.py
FredrikBakken: Thanks for the input, @jochenchrist! Changed the default value for `precision` to `38` now.

PR #451, opened 2024-10-06
feat: add support for `VarcharType` in spark converter

FredrikBakken:
- [ ] Tests pass
- [X] ruff format


jochenchrist (changes requested):
varchar as defined as an alias for string and text. Please don't use a special handling heare.

You would need to work with a config object `sparkType` to express an VarcharType to spark.

https://datacontract.com/#config-object

jochenchrist:
@FredrikBakken I will reject / close this PR.

It would be great if you could create a new PR that uses a config option https://datacontract.com/#config-object for spark specific types.

Issue #449, opened 2024-09-30
Iceberg Import and Export


paulcichonski:
[Apache Iceberg](https://iceberg.apache.org/) is an Open Table Format (OTF) that can be used for storing data in object storage with a well defined schema.

The Iceberg spec defines its own data schema representation (see [Table Spec](https://iceberg.apache.org/spec/)). Tha spec also defines a [json serialization of the schema](https://iceberg.apache.org/spec/#appendix-c-json-serialization) that can be used for schema management in the iceberg tables. Using the Iceberg json schema definition it is possible to call Iceberg Catalog APIs (popular implementations include Glue, Hive, REST, Dynamo, Sql) to [create tables](https://py.iceberg.apache.org/api/#create-a-table) or to [evolve the schema on existing tables](https://py.iceberg.apache.org/api/#schema-evolution). 

This issue is to request that the datacontract tool support an Iceberg schema export that adheres to the above json encoding. This way, models defined in a datacontract yml could be used for managing Iceberg tables. For example, the cli command could look like:

```bash
datacontract export --model orders --format iceberg https://datacontract.com/examples/orders-latest/datacontract.yaml > orders.json
```

Ideally it could support exporting all models at once, but the current Iceberg json definition seems to support one model per json document, so the above could be a good start.

simonharrer:
I second this. Feel free to create a pull request for this! :-)

paulcichonski:
Thanks, I'll try to get one put together either this week or next week.

simonharrer:
It might be a good idea to also build an import along the way because then you can have roundtrip tests (import - export - assert-unchanged, export - import - assert-unchanged) that cover a lot of functionality with only a few tests.

PR #452, opened 2024-10-06
Add support for importing from iceberg table definition

paulcichonski:
Import side of #449 .

I'm not going to have time for writing up the export side of this for a few days (at least) so just going to start with the import side and see what folks think. I think the export side is also a bit more involved since it is necessary to parse `definitions` into nested struct types of the iceberg schema. 

I've mostly copied the model of the Spark import for this since the Iceberg table format is quite similar to the Spark table definitions. Eventually it might be nice to add support for ingesting the table definition directly from the catalog (via the pyiceberg sdk), but I did not attempt to tackle that in this first iteration. I'm guessing I missed things so please just let me know what needs to be changed. 

This assumes the table definition is in JSON format as defined by https://iceberg.apache.org/spec/#appendix-c-json-serialization.

- [x] Tests pass
- [x] ruff format

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Added
+- Support for import from Iceberg table definitions.
+
 ## [0.10.13] - 2024-09-20
 
 ### Added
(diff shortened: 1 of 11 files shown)


paulcichonski:
re: tests, I get a lot of local test failures when running from the `main` branch so I haven't been able to confirm all tests are passing, but I don't think my changes should impact any tests other than the ones I added directly.

jochenchrist:
Thank you for your contribution. Looks good to me.

PR #469, opened 2024-10-18
Add support for trinoType field for custom Trino types

lucaslortiz:
# Description
This PR adds support for custom Trino types using `trinoType` field under `config` on the field definition.

# Issue
Trino has custom data types, for example `ROW` ([docs](https://trino.io/docs/current/language/types.html#row)), which were not supported by the default type definitions.
Similarly to what is used with other server types, I've added a custom field named `trinoType`, which enables any custom types to be used.

- [x] Tests pass
- [x] ruff format


jochenchrist:
Thanks for this focused PR.

PR #473, opened 2024-10-18
Fix issue with serializing from parquet to duckdb with timestamps

RobertLD:
- also add unit test to promote more tests be written around conversion (will add more later)

- [ ] Tests pass
- [x] ruff format


simonharrer:
See https://datacontract.com/#data-types which defines timestamp a synonym to timestamp_tz

RobertLD:
I think the crux of the issue is that duckdb defines timestamp and datetime
as synonyms. Differing from that definition. With convert to duckdb type
only defining a one-to-one mapping; how would be support the equivalent
aliases

datacontract/export/sql_type_converter.py
```diff
@@ -226,9 +226,10 @@ def convert_to_duckdb(field: Field) -> None | str:
         "bigint": "BIGINT",
         "date": "DATE",
         "time": "TIME",
-        "timestamp": "TIMESTAMP WITH TIME ZONE",
+        "datetime": "TIMESTAMP",
+        "timestamp": "TIMESTAMP",
```
jochenchrist, line 230: timestamp is a synonym with timestamp_tz and should have the same type

jochenchrist: plus `datetime` is not a valid Data Contract Data Typehttps://datacontract.com/#data-types

jochenchrist:
I made some adjustments to merge this PR.

PR #493, opened 2024-10-31
chore: use `enums` to ensure valid result values

FredrikBakken:
Using `enums` to ensure that the `result` values can only be set to any of the values defined in the `ResultEnum` class.

- [ ] Tests pass
- [X] ruff format


FredrikBakken:
Good to know, I'll keep that in mind if I look into contributing some more `enum` changes in the future :)

Issue #470, opened 2024-10-18
Add support for Importing datacontract from parquet


RobertLD:
It would be useful to import a stub datacontract file from the schema implemented in a parquet file

PR #494, opened 2024-11-01
Add parquet importer

stefannegele:
Closes https://github.com/datacontract/datacontract-cli/issues/470

- [x] Tests pass
- [x] ruff format


stefannegele:
Hmmm. Ruff complains here while not complaining in my local environment. Tried running `ruff format`/`ruff check` again. Maybe someone else has an idea what's wrong? Otherwise i will have to check again next week.

pyproject.toml
```diff
@@ -16,6 +16,7 @@ classifiers = [
 requires-python = ">=3.10"
 dependencies = [
   "typer>=0.12,<0.13",
+  "pyarrow>=12.0.0",
```
jochenchrist, line 19: how large is pyarrow?

stefannegele: Hmmm, quite big, about 100MB. I guess it brings some arrow binaries with it. Should i better make an extra (parquet) for it?

Issue #443, opened 2024-09-26
Enhance error reporting and handling for large JSON validations


NiklasA:
First of all, thank you very much for the great work and for providing the `datacontract-cli`!

----

While using it, we noticed three topics:

1. We use the `datacontract-cli` to check the structure and content of large JSON files. Currently, the `fastjsonschema.compile` process stops at the first error, which means that if there are multiple errors in the JSON file, the `datacontract-cli` has to be executed several times until all errors are fixed. We also considered executing the `data_contract.test()` for each object in the JSON file, but the overhead is too large, resulting in poor performance. Would it be possible to adjust the following call:

    ```python
    validate = fastjsonschema.compile(
        schema,
        formats={"uuid": r"^[0-9a-fA-F]{8}\b-[0-9a-fA-F]{4}\b-[0-9a-fA-F]{4}\b-[0-9a-fA-F]{4}\b-[0-9a-fA-F]{12}$"},
    )
    ```

    so that it works more like the following approach:

    ```python
    validate = fastjsonschema.compile(
        schema,
        formats={"uuid": r"^[0-9a-fA-F]{8}\b-[0-9a-fA-F]{4}\b-[0-9a-fA-F]{4}\b-[0-9a-fA-F]{4}\b-[0-9a-fA-F]{12}$"},
    )
    
    def validate_array(data):
        errors = []
        for index, item in enumerate(data):
            try:
                validate(item)
            except JsonSchemaException as e:
                # append error to list
                errors.append({"index": index, "error": str(e)})
        return errors
    ```

2. Additionally, we noticed that in `data_contract.py` on line 202, the model name is not included in the list of executed checks.

    ```python
    run.checks.append(
        Check(type=e.type, name=e.name, result=e.result, engine=e.engine, reason=e.reason, **model=e.model**)
    )
    ```

3. Lastly, it would be really helpful for debugging purposes if we could identify which JSON object in an array caused the error. Could the `validate_json_stream` method be extended so that it identifies the primary key of the related model in the data contract and returns the value from the found attribute along with the error if needed?

----

Let me know if I can help to create a pull request for those topics!

Best regards
Niklas

jochenchrist:
Hi @NiklasA,

Thanks for your feedback and these improvement ideas.

1. I understand the idea and it would certainly be doable. I hesitate a little, what happens when you have millions of JSON records that all have the same schema issue. Would you really like to see millions of errors in the response? We would need to implement a max number of errors (like 100?) here.

[...]

PR #505, opened 2024-11-17
Enhance error reporting and handling for large JSON validations

NiklasA:
Pull request for issue: **https://github.com/datacontract/datacontract-cli/issues/443**

- [x] Tests pass
- [x] ruff format

### DIFF
--- a/datacontract/engines/fastjsonschema/check_jsonschema.py
+++ b/datacontract/engines/fastjsonschema/check_jsonschema.py
@@ -2,7 +2,7 @@ import json
 import logging
 import os
 import threading
-from typing import List
+from typing import List, Optional
 
 import fastjsonschema
 from fastjsonschema import JsonSchemaValueException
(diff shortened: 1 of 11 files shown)


NiklasA:
@pascalrosenberger / @ManbirP could please review the PR, before I will change the status from draft to finish?

datacontract/engines/fastjsonschema/check_jsonschema.py
```diff
@@ -1,31 +1,112 @@
...
+            primary_key_field = field_name
+            break
+
+    # Cache the result.
+    with _cache_lock:
+        _primary_key_cache[model_name] = primary_key_field
+    return primary_key_field
+
+
+def get_primary_key_value(schema: dict, model_name: str, json_object: dict) -> str:
```
hirschsebastian, line 41: Since the return value can be theoretically None, I suggest to set as a return type Optional[str].

datacontract/engines/fastjsonschema/check_jsonschema.py
```diff
@@ -1,31 +1,112 @@
...
+        [
+            Check(
+                type=exception.type,
+                name=exception.name,
+                result=exception.result,
+                reason=exception.reason,
+                model=exception.model,
+                engine=exception.engine,
+                message=exception.message,
+            )
```
hirschsebastian, line 75: Since we use a property `message` here, I suggest to add the property to the Check class as Optional[str].

hirschsebastian (changes requested):
In three parts of the code I suggest adding Optional[str] to make the possibility of None more explicit.

PR #512, opened 2024-11-19
fix: json schema import with deeply nested non-required objects

DylanRodgers98:
- [x] Tests pass
- [x] ruff format

Importing JSON Schemas containing deeply nested objects without a `required` array currently fails with `KeyError: 'required'` due to indexed access of `property_schema["required"]`.

JSON Schema supports omitting `required` if all fields in an object are optional, so the importer should allow for this.

Note: top-level properties are not affected, this is only affecting nested objects.

Test and fixtures added, confirmed test fails (as described above) prior to changes.


jochenchrist:
Thx for your contribution!

PR #513, opened 2024-11-19
feat: add output option to datacontract import

DylanRodgers98:
- [x] Tests pass
- [x] ruff format

`datacontract export` supports `--output` option but `datacontract import` does not. When using the CLI via Docker this causes weird outputs when writing to file using `>` due to the 80 column limit _(FYI this can actually be increased with `docker run -e COLUMNS=n` but I think its good to have consistent API with `export`)_.

I have added a test to `tests/test_import_jsonschema.py` as JSON Schema is the domain I am currently working in, but I'm happy to add tests elsewhere if necessary.


jochenchrist:
Thanks for your contribution, this is super helpful!

PR #516, opened 2024-11-22
Changed input file detection to be recursive

fknierim:
This change enables catalog creation to include datacontracts in subfolders.

### DIFF
--- a/datacontract/cli.py
+++ b/datacontract/cli.py
@@ -346,7 +346,7 @@ def catalog(
     console.print(f"Created {output}")
 
     contracts = []
-    for file in Path().glob(files):
+    for file in Path().rglob(files):
         try:
             create_data_contract_html(contracts, file, path, schema)
         except Exception as e:


simonharrer:
One could already have it recursively if one used `**/*.yaml` as a global expression. Does one always want to use an rglob expression?

simonharrer:
It seems, powershell automatically expands "**/*.yaml" to "raw_zone\datacontract-testa.yaml raw_zone\datacontract-testb.yaml refined_zone\datacontract-testc.yaml trusted_zone\datacontract-testd.yaml trusted_zone\datacontract-teste.yaml" before invoking the command in the CLI. I am no powershell expert, but is there a way to prevent this?

jochenchrist:
I'd go with `rglob`.
Sounds reasonable that users want data contracts in any subdirectory.

simonharrer:
Ok, fine. :-)

PR #517, opened 2024-11-22
fix(export/bigquery): fixes schema issue with untyped objects in data contract

ricardolpd:
## Description:

Data models with no properties (untyped maps) currently create empty records which in turn create invalid bq schemas:

```
   someJson:
        type: object
        required: false
```
generate the following
```
{
        "name": "someJson:",
        "type": "RECORD",
        "mode": "NULLABLE",
        "description": null,
        "fields": []
      },
```

which causes the following error:
```
Error: googleapi: Error 400: Field rootObject.someJson is type RECORD but has no schema, invalid
│ 
```

But it should generate a `JSON` type instead:
```
{
    "name": "someJson",
     "type": "JSON",
     "mode": "NULLABLE",
     "description": null
}
```

This pr does a small fix to allow these types which can happen when having generic JSON maps

- [X] Tests pass
- [X] ruff format

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   - `engine` used to run checks
   - `sql_server_type` to define the type of SQL Server to use when engine is `sql`
 - Changelog support for `Info` and `Terms` blocks.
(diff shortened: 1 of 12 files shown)


jochenchrist:
Thanks for your contribution.

PR #526, opened 2024-11-29
escape e.g. double quotes from contents when rendering tag values

jens-gilion:
- [x] Tests pass
- [ ] ruff format

When content of contract contains double quotes, this closes the tag and the rest of the content overflows into the visible card itself.


jochenchrist:
Nice catch

jochenchrist:
Thanks for your contribution!

PR #532, opened 2024-12-05
Modify the arguments to narrow down the import target with `--dbt-model`

syou6162:
resolves https://github.com/datacontract/datacontract-cli/issues/533

During the refactoring of https://github.com/datacontract/datacontract-cli/pull/366, the names of the arguments were changed. As a result, even if `--dbt-model` is specified, all models are always output. This pull request fixes that.

Since changing the argument of `datacontract import` to `--dbt-nodes` would be a breaking change for end users, we only made the change in `datacontract/imports/dbt_importer.py`.

- [x] Tests pass
- [x] ruff format

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Changed
 
 ### Fixed
+- Modify the arguments to narrow down the import target with `--dbt-model` (#532)
 
 
 ## [0.10.15] - 2024-10-26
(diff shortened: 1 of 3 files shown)


simonharrer:
Good catch! Thanks for your help. I'll merge this.

simonharrer:
Hm, it seems, the tests fail. Can you have a look?

syou6162:
@simonharrer I'm sorry, I fixed it with [600545d](https://github.com/datacontract/datacontract-cli/pull/532/commits/600545dfd7562cf38686bd250170f2f9586dca07)!

PR #534, opened 2024-12-05
Type conversion when importing contracts into dbt and exporting contracts from dbt

syou6162:
## Problem
When importing/exporting to/from dbt, the conversion of column types is not taken into account.

When using BigQuery as an adapter for dbt, running `datacontract import --format dbt ...` will generate the following yaml.

```yaml
id: my-data-contract-id
info:
  title: my_data_model
  version: 0.0.1
  dbt_version: 1.8.8
models:
  my_model:
    description: ''
    fields:
      product_id:
        type: STRING
      quantity:
        type: INT64
    tags: []
```

`STRING` and `INT64` are [not defined as valid data types](https://datacontract.com/#data-types). In fact, the following error occurs when you try to export the imported yaml(`datacontract export --format dbt-sources ...`).

```
DataContractException: Run operation failed: [lint] Check that data contract 
YAML is valid - None - failed - 
data.models.my_data_model.fields.product_id.type must be 
one of ['number', 'decimal', 'numeric', 'int', 'integer', 'long', 'bigint', 
'float', 'double', 'string', 'text', 'varchar', 'boolean', 'timestamp', 
'timestamp_tz', 'timestamp_ntz', 'date', 'array', 'map', 'object', 'record', 
'struct', 'bytes', 'null'] - datacontract
```

## Solution
When importing dbt's yaml into a contract and exporting a contract to dbt, the type conversion has been modified to take into account.

---

- [x] Tests pass
- [x] ruff format

### DIFF
--- a/datacontract/export/dbt_converter.py
+++ b/datacontract/export/dbt_converter.py
@@ -88,7 +88,7 @@ def _to_dbt_source_table(model_key, model_value: Model, adapter_type: Optional[s
 
     if model_value.description is not None:
         dbt_model["description"] = model_value.description
-    columns = _to_columns(model_value.fields, False, False, adapter_type)
+    columns = _to_columns(model_value.fields, False, adapter_type)
     if columns:
         dbt_model["columns"] = columns
     return dbt_model
@@ -109,7 +109,7 @@ def _to_dbt_model(model_key, model_value: Model, data_contract_spec: DataContrac
         dbt_model["config"]["contract"] = {"enforced": True}
     if model_value.description is not None:
         dbt_model["description"] = model_value.description
-    columns = _to_columns(model_value.fields, _supports_constraints(model_type), True, None)
+    columns = _to_columns(model_value.fields, _supports_constraints(model_type), None)
     if columns:
         dbt_model["columns"] = columns
     return dbt_model
@@ -132,26 +132,25 @@ def _supports_constraints(model_type):
     return model_type == "table" or model_type == "incremental"
 
 
-def _to_columns(fields: Dict[str, Field], supports_constraints: bool, supports_datatype: bool, adapter_type: Optional[str]) -> list:
+def _to_columns(fields: Dict[str, Field], supports_constraints: bool, adapter_type: Optional[str]) -> list:
     columns = []
     for field_name, field in fields.items():
-        column = _to_column(field, supports_constraints, supports_datatype, adapter_type)
+        column = _to_column(field, supports_constraints, adapter_type)
         column["name"] = field_name
         columns.append(column)
     return columns
 
 
-def _to_column(field: Field, supports_constraints: bool, supports_datatype: bool, adapter_type: Optional[str]) -> dict:
+def _to_column(field: Field, supports_constraints: bool, adapter_type: Optional[str]) -> dict:
     column = {}
     adapter_type = adapter_type or "snowflake"
     dbt_type = convert_to_sql_type(field, adapter_type)
     if dbt_type is not None:
-        if supports_datatype:
-            column["data_type"] = dbt_type
-        else:
-            column.setdefault("tests", []).append(
-                {"dbt_expectations.dbt_expectations.expect_column_values_to_be_of_type": {"column_type": dbt_type}}
-            )
+        column["data_type"] = dbt_type
+    else:
+        column.setdefault("tests", []).append(
+            {"dbt_expectations.dbt_expectations.expect_column_values_to_be_of_type": {"column_type": dbt_type}}
+        )
     if field.description is not None:
         column["description"] = field.description
     if field.required:
--- a/datacontract/export/dbt_converter.py
+++ b/datacontract/export/dbt_converter.py
@@ -1,4 +1,4 @@
-from typing import Dict
+from typing import Dict, Optional
 
 import yaml
 
@@ -52,7 +52,7 @@ def to_dbt_staging_sql(data_contract_spec: DataContractSpecification, model_name
         # TODO escape SQL reserved key words, probably dependent on server type
         columns.append(field_name)
     return f"""
-    select 
+    select
         {", ".join(columns)}
     from {{{{ source('{id}', '{model_name}') }}}}
 """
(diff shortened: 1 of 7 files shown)


datacontract/imports/dbt_importer.py
```diff
@@ -58,7 +59,10 @@ def import_dbt_manifest(
     """
     data_contract_specification.info.title = manifest.metadata.project_name
     data_contract_specification.info.dbt_version = manifest.metadata.dbt_version
-
+    adapter_type = manifest.metadata.adapter_type
+    data_contract_specification.servers = {
```
syou6162, line 63: This information is useful as type conversion information when exporting contracts to dbt (and dbt-sources), so we have made it output.

This information is important when exchanging multiple dbt pipelines such as data mesh with data contracts.

simonharrer:
Awesome. Thanks for the contribution. Could you add a test for that, perhaps your example of importing from a dbt file that uses bigquery types?

datacontract/export/dbt_converter.py
```diff
@@ -130,25 +132,25 @@ def _supports_constraints(model_type):
     return model_type == "table" or model_type == "incremental"
 
 
-def _to_columns(fields: Dict[str, Field], supports_constraints: bool, supports_datatype: bool) -> list:
```
syou6162, line 133: The dbt supported by datacontract-cli is 1.8 or higher, and dbt has supported `data_type` since 1.5. Therefore, we decided that it would be fine to always set `supports_datatype` to True, and removed it from the arguments.

- https://github.com/datacontract/datacontract-cli/blob/main/pyproject.toml#L89-L91
- https://github.com/dbt-labs/dbt-core/blob/1.5.latest/CHANGELOG.md

syou6162:
@simonharrer I've added tests for each case of export and import! The difference in the fixture json file for testing is large, but I think it will be easier to understand if you look at the differences for [each commit](https://github.com/datacontract/datacontract-cli/pull/534/commits).

datacontract/export/dbt_converter.py
```diff
@@ -130,25 +132,25 @@ def _supports_constraints(model_type):
...
     if dbt_type is not None:
-        if supports_datatype:
-            column["data_type"] = dbt_type
-        else:
-            column.setdefault("tests", []).append(
-                {"dbt_expectations.dbt_expectations.expect_column_values_to_be_of_type": {"column_type": dbt_type}}
-            )
+        column["data_type"] = dbt_type
+    else:
+        column.setdefault("tests", []).append(
```
syou6162, line 151: We might as well delete this together.

datacontract/export/dbt_converter.py
simonharrer: It's not about the dbt version. 

We use the same code to generate the YAML for sources and for models. Sources do not support the datatype attribute, only models do. Can you revert, please?

syou6162: @simonharrer Thank you for your reply!

> Sources do not support the datatype attribute, only models do

I tried reading the source code for dbt-core. The definition for Sources is done in `SourceDefinition`, and the definition for columns is done in `ColumnInfo`. `data_type` is included in this.

- https://github.com/dbt-labs/dbt-core/blob/main/core/dbt/artifacts/resources/v1/source_definition.py
- https://github.com/dbt-labs/dbt-core/blob/main/core/dbt/artifacts/resources/v1/components.py#L65

In JSON Schema for the dbt yml file, `data_type` is also defined in `column_properties` within `columns`.

- https://github.com/dbt-labs/dbt-jsonschema/blob/main/schemas/latest/dbt_yml_files-latest.json#L903-L908
- https://github.com/dbt-labs/dbt-jsonschema/blob/main/schemas/latest/dbt_yml_files-latest.json#L1208-L1210

In fact, if you look at the contents generated by `manifest.json`, you can see that `data_type` is defined in Sources, so you can tell that Sources supports the `data_type` attribute.

```json
{
  "metadata": {
    "dbt_schema_version": "https://schemas.getdbt.com/dbt/manifest/v12.json",
    "dbt_version": "1.8.8",
    "generated_at": "2024-12-05T23:58:59.230157Z",
    "invocation_id": "41248e0c-2224-4a15-9c47-8b69f800bf76",
    "env": {},
    "project_name": "my_project",
    "project_id": "1ae1414bfc8e33f4f3905e27dbd8c461",
    "user_id": null,
    "send_anonymous_usage_stats": true,
    "adapter_type": "bigquery"
  },
  "nodes": {
    "source.my_project.my_dataset.my_table": {
      "database": "my_project",
      "schema": "my_dataset",
      "name": "my_table",
      "resource_type": "source",
      "package_name": "my_package",
      ...,
      "source_name": "my_project",
      "source_description": "",
      "loader": "",
      "identifier": "my_table",
      "quoting": {
        "database": null,
        "schema": null,
        "identifier": null,
        "column": null
      },
      "loaded_at_field": null,
      "freshness": null,
      "external": null,
      "description": "",
      "columns": {
        "shopId": {
          "name": "shopId",
          "description": "shop id",
          "meta": {},
          "data_type": "STRING",
          "constraints": [],
          "quote": null,
          "tags": []
        },
        "userId": {
          "name": "userId",
          "description": "user id",
          "meta": {},
          "data_type": "STRING",
          "constraints": [],
          "quote": null,
          "tags": []
        }
      },
      "meta": {
        "manual_data_usage": false
      },
      "source_meta": {},
      "tags": [],
      "config": {
        "enabled": true
      },
      "patch_path": null,
      "unrendered_config": {},
      ...,
      "created_at": 1731295492.503308
    }
  }
}
```

From the above survey, I don't think it is necessary to revert the relevant line, but what do you think, @simonharrer?

syou6162: So, does this mean that the conclusion here is that `data_type` is also supported by Sources, so there is no need to revert the corresponding line?

If so, please continue to review it. If there is anything that concerns you, please feel free to point it out!

datacontract/export/dbt_converter.py
simonharrer: Can be improved later

PR #541, opened 2024-12-10
Ensure 'name' is the first column when exporting in dbt format, considering column attributes

syou6162:
Currently, exporting in dbt format will output the following yaml(`datacontract export --format dbt-sources --output datacontract.yml input.yml`).

```yaml
version: 2
sources:
- name: my-data-contract-id
  tables:
  - name: my_product
    description: product table
    columns:
    - data_type: STRING
      description: product id
      tests:
        - not_null
        - unique
      meta:
        ...
      name: product_id
```

This yaml is valid as dbt yaml, but the column name is the last of the column attributes. Since the column name is the first attribute we want to see, this pull request modifies the column name so that it is the first of the column attributes.

- [x] Tests pass
- [x] ruff format

### DIFF
--- a/datacontract/export/dbt_converter.py
+++ b/datacontract/export/dbt_converter.py
@@ -135,14 +135,13 @@ def _supports_constraints(model_type):
 def _to_columns(fields: Dict[str, Field], supports_constraints: bool, adapter_type: Optional[str]) -> list:
     columns = []
     for field_name, field in fields.items():
-        column = _to_column(field, supports_constraints, adapter_type)
-        column["name"] = field_name
+        column = _to_column(field_name, field, supports_constraints, adapter_type)
         columns.append(column)
     return columns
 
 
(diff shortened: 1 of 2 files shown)


datacontract/export/dbt_converter.py
```diff
@@ -135,14 +135,13 @@ def _supports_constraints(model_type):
...
-        column["name"] = field_name
+        column = _to_column(field_name, field, supports_constraints, adapter_type)
         columns.append(column)
     return columns
 
 
-def _to_column(field: Field, supports_constraints: bool, adapter_type: Optional[str]) -> dict:
-    column = {}
+def _to_column(field_name: str, field: Field, supports_constraints: bool, adapter_type: Optional[str]) -> dict:
+    column = {"name": field_name}
```
syou6162, line 144: We do not need to fix the test case in this pull request, because python's dict does not consider the order when comparing.

- https://github.com/datacontract/datacontract-cli/blob/main/tests/test_export_dbt_sources.py#L134

PR #543, opened 2024-12-11
fix: populate database and schema values for bigquery in exported dbt sources

syou6162:
## Problem

Run the following command on the following yaml file:

```
datacontract export --format dbt-sources --output out.yml datacontract.yaml --server production
```

```yaml
id: my_contract
info:
  title: my_models
  version: 0.0.1
  dbt_version: 1.8.8
models:
  my_model:
    ...
servers:
  production:
    type: bigquery
    project: my-project
    dataset: my_dataset
```

This will result in a yaml file being generated with null database and schema values. Ideally, the BigQuery project name and dataset name should be entered here.

```yaml
version: 2
sources:
- name: my_models
  tables:
  - name: my_model
    ...
  database: null
  schema: null
```

## Solution
In the case of BigQuery, we added conditions so that project corresponds to database and dataset corresponds to schema.

---

- [x] Tests pass
- [x] ruff format

### DIFF
--- a/datacontract/export/dbt_converter.py
+++ b/datacontract/export/dbt_converter.py
@@ -72,8 +72,12 @@ def to_dbt_sources_yaml(data_contract_spec: DataContractSpecification, server: s
     adapter_type = None
     if found_server is not None:
         adapter_type = found_server.type
-        source["database"] = found_server.database
-        source["schema"] = found_server.schema_
+        if adapter_type == "bigquery":
+            source["database"] = found_server.project
+            source["schema"] = found_server.dataset
+        else:
+            source["database"] = found_server.database
+            source["schema"] = found_server.schema_
 
     for model_key, model_value in data_contract_spec.models.items():
         dbt_model = _to_dbt_source_table(model_key, model_value, adapter_type)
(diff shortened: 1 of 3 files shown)


datacontract/export/dbt_converter.py
```diff
@@ -72,9 +72,14 @@ def to_dbt_sources_yaml(data_contract_spec: DataContractSpecification, server: s
...
-        source["database"] = found_server.database
-        source["schema"] = found_server.schema_
+        if adapter_type == "bigquery":
+            source["database"] = found_server.project
+            source["schema"] = found_server.dataset
+        else:
+            source["database"] = found_server.database
+            source["schema"] = found_server.schema_
 
+    source["tables"] = []
```
syou6162, line 82: The purpose is the same as https://github.com/datacontract/datacontract-cli/pull/541. We want to display information such as which project or dataset the individual tables in the dbt source are in, and who the owner is, at the top of the yaml.

datacontract/export/dbt_converter.py
```diff
@@ -72,9 +72,14 @@ def to_dbt_sources_yaml(data_contract_spec: DataContractSpecification, server: s
     adapter_type = None
     if found_server is not None:
         adapter_type = found_server.type
-        source["database"] = found_server.database
-        source["schema"] = found_server.schema_
+        if adapter_type == "bigquery":
```
syou6162, line 75: The specifications for the BigQuery server type are described below.

- https://github.com/datacontract/datacontract-specification?tab=readme-ov-file#bigquery-server-object

PR #544, opened 2024-12-11
Fixing the options for importing and exporting to standard output

syou6162:
When executing the import and export subcommands, if `--output` is not specified, the data contract yaml will be output to standard output.

However, when outputting yaml to standard output, since [rich.console](https://rich.readthedocs.io/en/stable/console.html) is used, long descriptions in the yaml may be truncated. This is troublesome because when using file redirection, invalid yaml may be output. This is especially true if you are using a tool like [yq](https://github.com/mikefarah/yq).

Therefore, this pull request adds an option to avoid this.

- [x] Tests pass
- [x] ruff format


simonharrer:
Thanks for the contribution! :-)

Issue #542, opened 2024-12-10
When importing in dbt format, modify to import richer information.


syou6162:
The current `datacontract import --format dbt` only outputs simple information as a contract. It is only the column name / description / column type / tag.

- https://github.com/datacontract/datacontract-cli/blob/7d34626d673595294511464591705f92628a2e99/datacontract/imports/dbt_importer.py#L92-L96

However, dbt maintains rich information about columns, such as `relationships`, `primary_key`, and `not_null`, and this information should also be reflected in the contract.

syou6162:
There is a lot of information that can be reflected in the contract from the dbt yaml, so it's a lot to give at once, but I plan to send pull requests for this issue in several parts!

simonharrer:
Awesome. Looking forward to your help here!

PR #547, opened 2024-12-14
When importing in dbt format, add the dbt `not_null` information as a datacontract `required` field

syou6162:
partially resolve https://github.com/datacontract/datacontract-cli/issues/542

dbt can describe `not_null` information as `constraints` or `data_tests`.

However, this information is lost when `datacontract import --format dbt` is executed. This pull request improves this information to be reflected as the `required` field of the data contract.

- [x] Tests pass
- [x] ruff format

### DIFF
--- a/datacontract/imports/dbt_importer.py
+++ b/datacontract/imports/dbt_importer.py
@@ -78,12 +78,18 @@ def import_dbt_manifest(
         dc_model = Model(
             description=model_contents.description,
             tags=model_contents.tags,
-            fields=create_fields(manifest, model_unique_id=model_contents.unique_id, columns=model_contents.columns, adapter_type=adapter_type),
+            fields=create_fields(
+                manifest,
+                model_unique_id=model_contents.unique_id,
+                columns=model_contents.columns,
+                adapter_type=adapter_type,
+            ),
         )
 
         data_contract_specification.models[model_contents.name] = dc_model
 
     return data_contract_specification
+
 
 def convert_data_type_by_adapter_type(data_type: str, adapter_type: str) -> str:
     if adapter_type == "bigquery":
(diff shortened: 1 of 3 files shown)


datacontract/imports/dbt_importer.py
```diff
@@ -44,7 +46,9 @@ def read_dbt_manifest(manifest_path: str) -> Manifest:
     """Read a manifest from file."""
     with open(file=manifest_path, mode="r", encoding="utf-8") as f:
         manifest_dict: dict = json.load(f)
-    return Manifest.from_dict(manifest_dict)
+    manifest = Manifest.from_dict(manifest_dict)
+    manifest.build_parent_and_child_maps()
```
syou6162, line 50: We need to build child_map in advance for this line

- https://github.com/datacontract/datacontract-cli/pull/547/files#diff-b5d431da7ac31000c438134a0a2bc8d45257096a00e1144192579b244ca340d6R114

datacontract/imports/dbt_importer.py
```diff
@@ -74,27 +78,79 @@ def import_dbt_manifest(
...
+                "args": test_node.test_metadata.kwargs,
+            }
         )
-        for column in columns.values()
-    }
+    return column_tests
 
-    return fields
+
+def create_field(manifest: Manifest, model_unique_id: str, column: ColumnInfo, adapter_type: str) -> Field:
```
syou6162, line 138: `ColumnInfo` does not contain information about `data_tests`. Instead, `data_tests` information is contained in `Manifest`, and the `get_column_tests` function is used to determine whether the column contains a `not_null` test, and `required` is assigned.

- https://github.com/dbt-labs/dbt-core/blob/7df04b0fe461dcad3c2abcb67004178efaf445ce/core/dbt/artifacts/resources/v1/components.py#L58-L70

datacontract/imports/dbt_importer.py
```diff
@@ -74,27 +78,81 @@ def import_dbt_manifest(
...
+        if not test_node or test_node.resource_type != "test":
+            continue
+
+        if not isinstance(test_node, GenericTestNode):
+            continue
+
+        if test_node.column_name != column_name:
+            continue
+
+        if test_node.config.where is not None:
```
syou6162, line 127: In tests for dbt that include `not_null`, you can use the `where` keyword to add conditions. When comparing the `required` and `not_null` [specifications of datacontract](https://datacontract.com/), we decided to exclude the `where` keyword from the current specification, as it is difficult to describe the `required` condition in datacontract when the `where` keyword is specified.

- https://docs.getdbt.com/reference/resource-configs/where

PR #548, opened 2024-12-14
nit: Rename dbt's `tests` to `data_tests`

syou6162:
From dbt v1.8, `tests` are now called `data tests` to disambiguate from unit tests. Since it is now recommended to use `data_tests`, we will modify the export to dbt format to use `data_tests`.

- [x] Tests pass
- [x] ruff format


simonharrer:
Hm, dbt 1.8 was only released in May 2024. Do we need a `--dbt-version` parameter to offer a downgrade path in a graceperiod?

PR #549, opened 2024-12-16
Add Iceberg Export

### LINKED ISSUE #449: Iceberg Import and Export

paulcichonski:
- [ ] Tests pass
- [x] ruff format

Closes #449 

Took me longer than I was hoping it to, but I think this is a decent start at an Iceberg export. Happy to change it as needed though.


jochenchrist:
Can you check why the tests fail?

PR #567, opened 2024-12-24
Update README.md

Lucifer-02:
- [ x] Tests pass
- [ x] ruff format
- [ x] README.md updated (if relevant)
- [x ] CHANGELOG.md entry added

fix typos


pierre-monnet:
`odcs` and `odcs_v3` match to the same exporter. The v3 is the current version and so the default.
https://github.com/datacontract/datacontract-cli/blob/ff4813c20e8fc33cd896cb13cd5643c4788ff09f/datacontract/export/exporter_factory.py#L106-L112

jochenchrist:
`odcs` is an alias for `odcs_v3` and should always point to the latest version (we'll drop support for v2 soon)

PR #569, opened 2024-12-29
When exporting in dbt format, add the datacontract `references` field as a dbt `relationships` test

syou6162:
partially resolve https://github.com/datacontract/datacontract-cli/issues/542

datacontract can describe information about foreign keys in `references`. Also, dbt can describe tests for foreign keys in `relationships`, so I've made a modification so that when `datacontract export --format dbt-sources` is run, it will add to the dbt yaml as a `relationships` test.

If the direction of this pull request is fine, I will send a similar improvement pull request for `datacontract import --format dbt` as well.

- [x] Tests pass
- [x] ruff format


datacontract/export/dbt_converter.py
```diff
@@ -239,9 +246,21 @@ def _to_column(field_name: str, field: Field, supports_constraints: bool, adapte
...
+            column["data_tests"].append(
+                {
+                    "relationships": {
+                        "to": f"""source("{ref_source_name}", "{table_name}")""",
+                        "field": f"{column_name}",
+                    }
+                }
+            )
 
-        if not column["data_tests"]:
```
syou6162, line 243: I made this correction, but the scope in which this processing was performed was incorrect 🙏 

- https://github.com/datacontract/datacontract-cli/pull/548/files#diff-b03f6e792b001775280d2 e272e3faef45c6880886f91c19597a23662b7f574a3R242-R244

PR #570, opened 2024-12-29
When importing in dbt format, add the dbt `relationships` test field as a `references` in the data contract

syou6162:
partially resolve https://github.com/datacontract/datacontract-cli/issues/542, and related to https://github.com/datacontract/datacontract-cli/pull/569.

dbt can describe information about foreign keys as tests in `relationships`. Also, datacontract can describe information about foreign keys in references, so I've made a modification so that when `datacontract import --format dbt --source target/manifest.json` is run, it will add to the data contract as a `references` field

- [x] Tests pass
- [x] ruff format


syou6162:
The CI test has failed, but this is due to a fix on the main branch. Once the main branch test passes, this branch test should also pass.

- ref: https://github.com/datacontract/datacontract-cli/issues/575

PR #589, opened 2025-01-16
Allow datacontract import to be run offline

glM26:
Right now, schema and template are always downloaded from the default URLs, which might not be accessible in a restricted environment.
Add --schema and --template parameters to datacontract import.
- [ ] Tests pass
- [ ] ruff format


jochenchrist:
LGTM

PR #598, opened 2025-01-18
Ship spec jsons within the CLI.

### LINKED ISSUE #580: Ship spec jsons within the CLI
Include the json schemas used for schema validation within the CLI.

jochenchrist:
Resolves #580

- [x] Tests pass
- [x] ruff format


simonharrer:
I think it is okay to leave out the older schemas as it is fast moving and the schemas should be backward compatible. We should mention that the CLI tool no longer needs to download files when being run.

- At the moment, it never validates ODCS as always DCS is loaded. We should handle that.
- We should also include the init template (https://datacontract.com/datacontract.init.yaml) and ship it from within the CLI.

simonharrer:
LGTM

PR #612, opened 2025-01-27
fix: schema yaml file not included in docker build

DylanRodgers98:
Running the cli in Docker currently fails because schemas/datacontract-1.1.0.init.yaml is not included in the build.
Also, passing `--template` option currently fails due to an erroneous `if template is None` check.

- [x] Tests pass
- [x] ruff format


datacontract/data_contract.py
```diff
@@ -75,9 +75,8 @@ def __init__(
         }
 
     @classmethod
-    def init(cls, template: str, schema: typing.Optional[str] = None) -> DataContractSpecification:
-        if template is None:
-            template_str = get_init_template(template)
+    def init(cls, template: typing.Optional[str], schema: typing.Optional[str] = None) -> DataContractSpecification:
+        template_str = get_init_template(template)
```
DylanRodgers98, line 79: `get_init_template` handles when `template` is `None` and grabs the default template

PR #615, opened 2025-01-27
Adding new enum members should not be breaking change

DylanRodgers98:
Adding new members to enums without removing existing members should not be a breaking change. This PR adds a new `field_enum_members_added` breaking change rule with severity `info` - i.e. not breaking.

- [x] Tests pass
- [x] ruff format


jochenchrist:
I am not so sure here.
Adding new and unexpected enum values are typically a major change that consumers must be a aware of and adopt their processes/code, e.g. in switch matching statements.

DylanRodgers98:
> e.g. in switch matching statements.

Ah that's a great point

PR #648, opened 2025-02-19
Pass csv parameters during duckdb connection

Halpph:
- [ ] Tests pass
- [ ] ruff format

### DIFF
--- a/datacontract/model/data_contract_specification.py
+++ b/datacontract/model/data_contract_specification.py
@@ -58,6 +58,9 @@ class Server(pyd.BaseModel):
     dataset: str | None = None
     path: str | None = None
     delimiter: str | None = None
+    header: bool | None = None
+    escape: str | None = None
+    encoding: str | None = None
     endpointUrl: str | None = None
     location: str | None = None
     account: str | None = None
--- a/datacontract/engines/soda/connections/duckdb.py
+++ b/datacontract/engines/soda/connections/duckdb.py
@@ -42,14 +42,29 @@ def get_duckdb_connection(data_contract, server, run: Run):
         elif server.format == "csv":
             columns = to_csv_types(model)
             run.log_info("Using columns: " + str(columns))
-            if columns is None:
-                con.sql(
-                    f"""CREATE VIEW "{model_name}" AS SELECT * FROM read_csv('{model_path}', hive_partitioning=1);"""
-                )
-            else:
-                con.sql(
-                    f"""CREATE VIEW "{model_name}" AS SELECT * FROM read_csv('{model_path}', hive_partitioning=1, columns={columns});"""
-                )
+
+            # Start with the required parameter.
+            params = ["hive_partitioning=1"]
+
+            # Loop over optional CSV parameters.
+            for param in ("delimiter", "header", "escape", "encoding"):
+                value = getattr(server, param)
+                if value is not None:
+                    # Wrap string values in quotes.
+                    if isinstance(value, str):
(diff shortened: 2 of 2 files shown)


datacontract/model/data_contract_specification.py
```diff
@@ -58,6 +58,15 @@ class Server(pyd.BaseModel):
     dataset: str | None = None
     path: str | None = None
     delimiter: str | None = None
+    header: bool | None = None
+    escape: str | None = None
+    all_varchar: bool | None = None
```
jochenchrist, line 63: please use lowerCamelCase

datacontract/schemas/datacontract-1.1.0.schema.json
```diff
@@ -1212,11 +1212,48 @@
...
           "type": "string",
-          "enum": [
-            "new_line",
-            "array"
+          "anyOf": [
+            { "enum": ["new_line", "array"] },
+            { "pattern": "^.$" }
           ],
-          "description": "Only for format = json. How multiple json documents are delimited within one file"
+          "description": "For JSON format, only 'new_line' or 'array' is allowed to indicate how multiple JSON documents are delimited. For CSV format, any single character can be used as the delimiter between columns. Only valid for CSV."
```
jochenchrist, line 1219: the master for the schema is here: https://github.com/datacontract/datacontract-specification/

Halpph: @jochenchrist should I not modify this one and just change the one in the other repo then?

pyproject.toml
```diff
@@ -26,6 +26,7 @@ dependencies = [
   "rich>=13.7,<13.10",
   "sqlglot>=26.6.0,<27.0.0",
   "duckdb==1.1.2",
+  "fsspec",
```
jochenchrist, line 29: Where is this used?

datacontract/schemas/datacontract-1.1.0.schema.json
jochenchrist: Please don't the file here, but propose a change in the other repo.

We still can continue to use the provided keys in the data_contract_specification (as custom extensions)

tests/test_duckdb_csv.py
jochenchrist, line 1: The filename is not in line with the naming conventions (test_<command>_<servertype>.py
Please move these tests to test_test_clocal_csv.

pyproject.toml
stefanedwards: In file `datacontract\engines\soda\connections\duckdb.py`, the method `sniff_csv_header` uses `duckdb.from_csv_auto` to read a csv file as a stream.

Without `fsspec`, the following fails:

```
return duckdb.from_csv_auto(io.BytesIO(header_line), **csv_params).columns
E       duckdb.duckdb.InvalidInputException: Invalid Input Error: This operation could not be completed because required module 'fsspec' is not installed
```

tests/test_duckdb_csv.py
stefanedwards: But the tests do not test the test routine. They test whether duckdb can handle mismatched column specifications when reading csv-files.

Do you still want them in `test_test_local_csv.py`?

dmaresma:
Hi there is also a regression, introduce by the sniff_csv_header, 

when using a `with open(model_path, 'rb') ` the model_path could be an abfss:// , s3://  blob or datalake store url, that is not supported by the python io `open()` function.

PR #668, opened 2025-02-28
Add DBT sql dialect column data types with support for optional --server flag

skidave:
- [x] Tests pass
- [x] ruff format

## Changes

-  `dbt` & `dbt-sources` export formats now support the optional `--server` flag to adapt the DBT column `data_type` to specific SQL dialects
- Fix multiline descriptions in the DBT export functionality

Motivation behind this PR is that we use DBT and BigQuery as part of ETL pipelines, but when exporting datacontracts into DBT yaml formats, the column `data_types` didn't match those expected in Bigquery.

The DBT exporter now relies on the `server`/`adapter_type` being passed down to the `_to_column()` function to correctly determine the column data type.


jochenchrist:
LGTM, thanks for your contribution.

PR #681, opened 2025-03-07
deep nesting of json objects in duckdb

stefanedwards:
- [ x ] Tests pass
- [ ] ruff format

When importing a json file with nested objects into duckdb, duckdb makes some educated guesses on the data types of the nested objects. But when it fails, there is no recourse by specifying the nested data model in the datacontract, as the model's fields are entirely ignored when reading the json-file.

This fix provides a solution that allows a complete, nested data structure to be used when importing the json file into duckdb.


datacontract/export/json_type_converter.py
jochenchrist, line 1: This is a very generic name for this file.

stefanedwards: Just following the convention, cf. [`csv_type_converter.py`](https://github.com/datacontract/datacontract-cli/blob/77d12b3724a7fe515aeeb97336d70198d3905d57/datacontract/export/csv_type_converter.py). But I'd happily combine the two files into a `duckdb_type_converter(s)`.

PR #697, opened 2025-03-19
Better ODCS Support

simonharrer:
- [ ] Tests pass
- [ ] ruff format


datacontract/export/odcs_v3_exporter.py
```diff
@@ -5,6 +5,17 @@
 from datacontract.export.exporter import Exporter
 from datacontract.model.data_contract_specification import DataContractSpecification, Field, Model
 
+yaml.SafeDumper.org_represent_str = yaml.SafeDumper.represent_str
+
+
+def repr_str(dumper, data):
```
jochenchrist, line 11: Why is this function on top and not part of the class?

Also, the name is not clear to me.

PR #743, opened 2025-05-02
Add example for testing with local files

mitchtech:
I was wrongly using the S3 sample for reference and using `location:` instead of `path:` for the local file location. Hopefully this example will help future users.


jochenchrist:
Thanks for your contribution!

PR #748, opened 2025-05-09
Change behaviour of DBT constraints vs data tests when no model type specified on contract

skidave:
- [x] Tests pass
- [x] ruff format

The [default DBT behaviour is to materialize models as views](https://docs.getdbt.com/docs/build/materializations#:~:text=By%20default%2C%20dbt%20models%20are%20materialized%20as%20%22views%22.) which do not support `constraints`, only `data tests`.

However, we don't want to assume a default DBT model type if none is specified in the contract and so the DBT exporter defaults to the more permissive `data tests` rather than `constraints`, and does not include the `models[].config.materialized` key in the DBT output.

This has *no* impact on data contracts that specify the `model.type` in the model object.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## Unreleased
 
+### Changed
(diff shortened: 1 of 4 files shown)


jochenchrist:
LGTM, thanks for the contribution!

PR #758, opened 2025-05-16
Add support for Variant type

robert-altmiller:
We are adding support for the variant type because its a new datatype which is currently in public preview:
https://docs.databricks.com/aws/en/sql/language-manual/data-types/variant-type

With these changes in the PR and adding the following two settings in the Databricks notebook or Python file Databricks variant types are completely supported and also deployable using DDLs created from data_contral.export("sql").

1. spark.conf.set("spark.databricks.delta.schema.typeCheck.enabled", "false")
2. from datacontract.model import data_contract_specification
data_contract_specification.DATACONTRACT_TYPES.append("variant")

When I create the contract with these setting above and changes to your code base below Variant types are stored in the ODCS Data Contract in the following way:

image

When I generate the SQL DDLs to deploy into a Databricks Catalog.Schema they come out the following way:

image

This SQL DDL is deployable above in Databricks as long as the spark.conf() setting mentioned earlier is set.

### DIFF
--- a/datacontract/export/odcs_v3_exporter.py
+++ b/datacontract/export/odcs_v3_exporter.py
@@ -202,6 +202,8 @@ def to_logical_type(type: str) -> str | None:
         return "array"
     if type.lower() in ["array"]:
         return "array"
+    if type.lower() in ["variant"]:
+        return "variant"
     if type.lower() in ["null"]:
         return None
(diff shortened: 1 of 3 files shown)


jochenchrist (approved):
LGTM

PR #760, opened 2025-05-17
Complex Types Fix When Exporting to SQL and 'databricksType' is Defined

robert-altmiller:
These changes add physical type back into the ODCS contract, and allow us to capture the full complex nested datatype (e.g. array, object, record, struct) --> 

image

With these changes here is what this sql block above translates into the ODCS Data Contract prior to being converted to sql DDL:

image

image


jochenchrist (approved):
LGTM

PR #768, opened 2025-05-21
fix: Soda Exporter did not correctly pass server object to create_checks function

Christopher-Lawford:
The soda exporter create_checks function expects to receive a server object but only receives a server_name string.
This fix adds in a step to get the server object from the data contract using the get_server function from the data_contract_test module.

Example:
If you run the exporter like so 'datacontract export .\src\\datacontract.yaml --format sodacl --server dev'  with a data contract like:
```

id: my-data-contract-id
info:
  title: Transactions
  version: 1.0.0
  description: Transactions from datahub since 2020-01-01

servers:
  dev:
    type: databricks
    catalog: core_dev
    schema: silver

models:
  initial_data:
    type: table
    description: Initial data for testing
    fields:
      id:
        type: integer
        description: ID of the person
      name:
        type: string
        description: Name of the person
        config:
          databricksType: decimal(10, 2)
      added_timestamp:
        type: timestamp
        description: Timestamp when the data was added

```

I would expect this to work but it fails with 'AttributeError: 'str' object has no attribute 'type'.
image

- [ ] Tests pass
- [ ] ruff format

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - `datacontract import --format spark`: Added support for spark importer table level comments (#761)
 
+### Fixed
+
+- `datacontract export --format sodacl`: Fix resolving server when using `--server` flag (#768)
+
 ## [0.10.26] - 2025-05-16
 
 ### Changed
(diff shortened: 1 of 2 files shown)


jochenchrist:
Good catch! Thanks for reporting!

Issue #737, opened 2025-04-24
Fix Bug:  Differentiate between the Nested and Repeated fields for bigquery when executing datacontract import.


gregoireAGBG:
Hello, 
So in version of datacontract-cli 0.10.24 and preivous versions, there is type mismatch in the schema that is being imported from bigquery for **REPEATED RECORD**:

when executing 
`datacontract import --format biquery --biquery-project my_gcp_project --biquery-dataset my_dataset --bigquery-table my_table >> dc_my_gcp_project_my_dataset_my_table.yml `

inside my dc_my_gcp_project_my_dataset_my_table.yml the following will be generated for the my_bq_col (nested : 

`      my_bq_col:
        **type: object
        required: false
        fields**:
          subfield1:
            type: string
            required: true
          subfiel2:
            type: string
            required: false
          subfield3:
            type: string
            required: false`

After adding the server informations, a manual modification is needed in order to be able to execute : 
`datacontract test dc_my_gcp_project_my_dataset_my_table.yml --logs`

the correct adaptation is as follow: 
`      my_bq_col:
        **type: array
        items:
          type: record
          fields:**
            subfield1:
              type: string
              required: true
            subfield2:
              type: string
              required: false
            subfield3:
              type: string
              required: false`

Below you can find the ouput of executing `datacontract test dc_my_gcp_project_my_dataset_my_table.yml --logs`
Before the modification: 
descriptions Check that field my_bq_col has type STRUCT<subfield1 STRING, subfield2 STRING, subfield3 STRING>: Type Mismatch, Expected Type: STRUCT<subfield1 STRING, subfield2 STRING, subfield3 STRING>;
Actual Type: ARRAY<STRUCT<subfield1 STRING, subfield2 STRING, subfield3 STRING>>

After the modification: 

passed │ Check that field my_bq_col has type ARRAY<STRUCT<subfield1 STRING, subfield2 STRING, subfield3 STRING>>│ my_bq_col│         │

Explanation: 
- So currently datacontract-cli **import --format biquery** is handling the nested repeated fields as nested fields : 
Nested fields (or Struct ) are defined by:  type": "RECORD", "mode": "NULLABLE",
while nested repeated fields (or Array<Struct>) are defined by: type": "RECORD", "mode": "REPEATED",
RECORD type seems to be handled as a unique case and not have different behavior depending on the mode.

To resolve the issue you should check the mode of the field and handle the mode "REPEATED" as described above to enable a valid datacontract schema check. 

[...]

PR #769, opened 2025-05-21
resolve issue #737 fix type conversion issue for bigquery target 

codeagent-art:
… doing datacontract import followed by datacontract test after adding server information

- [ ] Tests pass
- [ ] ruff format


datacontract/export/bigquery_converter.py
```diff
@@ -98,10 +98,12 @@ def map_type_to_bigquery(field: Field) -> str:
         return "FLOAT64"
     elif field_type.lower() == "boolean":
         return "BOOL"
-    elif field_type.lower() in ["timestamp", "timestamp_tz"]:
+    elif field_type.lower() in "timestamp":
         return "TIMESTAMP"
     elif field_type.lower() == "date":
         return "DATE"
+    elif field_type.lower() ==  "timestamp_tz":
+        return "TIME"
```
simonharrer, line 106: this is not correct, I think. See https://datacontract.com/#data-types timestamp and timestamp_tz are the same (basically alias for each other).

We do not have a time datatype. We could add one to the spec. Until then, we would need to resort to the config options.

datacontract/imports/bigquery_importer.py
```diff
@@ -167,13 +185,13 @@ def import_table_fields(table_fields):
...
 
-def map_type_from_bigquery(bigquery_type_str: str):
+def map_type_from_bigquery(bigquery_type_str: str, in_array=False):
     if bigquery_type_str == "STRING":
         return "string"
     elif bigquery_type_str == "BYTES":
         return "bytes"
     elif bigquery_type_str == "INTEGER":
-        return "int"
+        return "bigint" if in_array else "int"
```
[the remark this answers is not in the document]
gregoire-lab: Hi so regarding that INT64 type this is something that is internal to bigquery where the data_type of bq table within a struct or an array is represented as an INT64 and not an INTEGER. 

1. create or replace table `project_id.dataset_id.simple` as (
  select * from
  (select [1,2,3] as array_int64)
  left join
  (select  struct(1 as elem) as struct_int64)
  on true
)
2. when looking at the table you see the fields as INTEGER
Capture d'écran 2025-05-21 131625

3. But when looking at the information_schema: SELECT * FROM `project_id.dataset_id.INFORMATION_SCHEMA.COLUMNS` where table_name ='simple'  
you see that the bigquery internal representation is an INT64
Capture d'écran 2025-05-21 131420

datacontract/export/bigquery_converter.py
codeagent-art: Hello ok i have rollbacked that part to your original mapping. 
Would be great to add time to the list to arrive at a more complete list. 
As far as completeness, would be great to achieve a complete type coverage and most importantly a valid full cercle (meaning when importing a table from bigquery i aspect the datacontract test to work after adding the server informations) and i aspect the datacontract import to stay stable and not be changed too much. 

We are currently automatising the generation of newer datacontracts per table based on import from source functionnality (comparing models section of datacontracts to determine if the schema has evolved or not). We would like to not have to generate unnecessary datacontracts due to underlying changes in the way types are mapped or not. 

Would be great to include tests to ensure that an imported datacontract when refined (adding valid server information) has a valid execution.

PR #771, opened 2025-05-25
Added support to spark importer to be able to import Spark dataframe object

robert-altmiller:
I have added support to be able to import a Spark dataframe object direct using the following methods:

image

Description is optional, and dataframe is that actual Spark dataframe object.  These new methods above only work for a single dataframe and single source.

Readme.md has been updated to reflect changes to spark importer, and export as source sql.  I have also updated a new user_datacontract.yml in tests.fixtures.import directory.  Updated the spark import test too.

image


README.md
```diff
@@ -1375,14 +1395,21 @@ datacontract import --format glue --source <database_name>
...
 
 ```bash
+# Example: Import Spark table(s) from Spark context
 datacontract import --format spark --source "users,orders"
 ```
 
+Another method to run importing from spark is to specify a keyword argument `dataframe` that is equal to the Spark dataframe object.  `Source` is equivalent to the table name.  The table `description` keyword argument is optional. This method **ONLY** works with a single `source` and `dataframe`.
+
+```bash
+# Example: Import Spark dataframe from Spark context
```
jochenchrist, line 1408: I would just show the programmatic example. 
I think 99% of users would use the python version.

robert-altmiller: I have modified this.

jochenchrist (approved):
LGTM

PR #785, opened 2025-06-05
add option to pass extra arguments, some small fixes

### LINKED ISSUE #782: Serving API behind a proxy: Fetch error response status is 404 /openapi.json
We are trying to serve the Datacontract API in a Kubernetes cluster on the following URL: `<root_url>/datacontract`. On the cluster, Traefik is used for networking. I have added the Docker container in a deployment, and added a service, an ingressRoute and Middleware to send the requests to the API. The middleware strips the prefix `/datacontract`. As a result, the endpoints `test`, `lint` and `export` are available. 
However, there is a[ know issue](https://fastapi.tiangolo.com/advanced/behind-a-proxy/) in FastAPI where the docs don't work in this setup. Since there is no way currently to set the `root_path` for FastAPI in the Dockerfile, what is the proposed solution? 
I can open a PR to pass `**kwargs` to the API, which would make it possible to start the API with `datacontract api --host 0.0.0.0 --port 4242 --root_path /datacontract`, but I don't know if you want to support this scenario.

bartzwemmer:
- [x] Tests pass
- [x] ruff format
Tests for the API pass. There are many unrelated tests with issues. Possibly caused by the umlaut in the test data. 
Fixes #782

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,7 +14,6 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 - `pytest tests\test_api.py`: Fixed an issue where special characters were not read correctly from file.
-- `./lint/resolve/py` Pydantic Deprecation warning: fixed by using the Class in stead of the instance on line 97
 
 ## [0.10.28] - 2025-06-05
 


datacontract/lint/resolve.py
```diff
@@ -130,7 +130,7 @@ def inline_definition_into_field(field, spec):
         return
 
     definition = _resolve_definition_ref(field.ref, spec)
-    for field_name in field.model_fields.keys():
+    for field_name in type(field).model_fields.keys():
```
simonharrer: why do we need this here? below, with definition, is it not necessary?

bartzwemmer: This is to suppress a future deprecation warning from pydantic. I can reproduce it in the morning.
They want to get the keys from the Class in stead of the Instance in the future. 
I discovered this while running pytest on the test_api.py

simonharrer: Can you move that into a separate PR, please? Because when switching to new behavior, it is unclear whether it still works for the minimum version of pydantic that we enforce.

datacontract/cli.py
```diff
@@ -488,14 +489,28 @@ def api(
...
+    # Default Datacontract uvicorn arguments
+    default_args = {
+        "app": "datacontract.api:app",
+        "port": port,
+        "host": host,
+        "reload": True,
+        "log_config": log_config,
+    }
+    # Add extra arguments to the default args, remove the leading --
+    trimmed_keys = list(map(lambda x : str(x).replace("--", ""),ctx.args[::2]))
```
jochenchrist: O, a lot of magic is happening here. Can you simplify the code here? E.g. by extracting  to a function?

datacontract/lint/resolve.py
bartzwemmer: I reverted it. I can open another PR next week for this, I think. Then I'll check the entire code base for this behavior, to make it uniform.

datacontract/cli.py
bartzwemmer: Did it

PR #792, opened 2025-06-11
Added support for Variant with Spark exporter, data_contract.test(), and import as source unity catalog

robert-altmiller:
Added support for Variant type with spark exporter, uc importer, and data_contract.test().  In the import_as_source() using unity catalog the Variant type did not convert over correctly or got ignored entirely.


robert-altmiller:
image

robert-altmiller:
image

robert-altmiller:
image

robert-altmiller:
image

datacontract/export/spark_converter.py
```diff
@@ -150,7 +152,7 @@ def to_spark_data_type(field: Field) -> types.DataType:
         return types.DateType()
     if field_type == "bytes":
         return types.BinaryType()
-    return types.BinaryType()
+    return types.StringType() # default if no condition is met
```
simonharrer, line 155: why change the default here? What's the reasoning?

robert-altmiller: Not all unknown types should default to BinaryType() — that is often too low-level.  I figured it is more user-friendly to default to StringType() if you don't recognize the field type.  Many tools (Delta Lake, Hive Metastore, JDBC clients) handle StringType gracefully but may not handle BinaryType well unless explicitly expected.

robert-altmiller: Why was BinaryType() chosen as the fallback?  Maybe I am missing something in my understanding.

jochenchrist: Binary as default seems wrong anyway.
String seems to be a safer choice.

datacontract/schemas/datacontract-1.1.0.schema.json
```diff
@@ -1097,6 +1097,7 @@
         "object",
         "record",
         "struct",
+        "variant",
```
jochenchrist: There is no variant data type: https://datacontract.com/#data-types

jochenchrist (changes requested):
We cannot just add a new data type to Data Contract Specification and its schema...
I might have missed this in an earlier PR, but this would violate the spec.

datacontract/export/spark_converter.py
robert-altmiller: This is what can happen if you keep it as binarytype()

image

datacontract/schemas/datacontract-1.1.0.schema.json
robert-altmiller: @jochenchrist how would we account for this because using the CLI method will fail if Variant is used as column types?  Or do we just say that is how it is supposed to work so you have to go use Python directly?  I don't get this error if I am using Python to interact with the data contract CLI.

image

datacontract/export/spark_converter.py
simonharrer: Fine by me. :-)

jochenchrist:
Data Contract Specification 1.2.0 has been released and it now supports data type `variant`. Finally, you can remove the workaround.

jochenchrist:
LGTM!

PR #793, opened 2025-06-12
fix: s3 url style path

stephanheinze:
the s3 url style "path" should not be bound to an endpoint starting with "http://". local min.io could be protected with ssl/tls as well.

- [ ] Tests pass
- [ ] ruff format


jochenchrist:
Thanks for this improvement!

PR #815, opened 2025-07-04
feat(odcs custom export)

### LINKED ISSUE #814: Support custom export for ODCS
Hello,

We would like to generate custom exports directly from ODCS. 

Currently, --format custom flag calls custom_export.py that only deals with DCS. 
#783 introduced html export using an odcs-specific default template.

To export using a custom template with DCS, user need to provide the following : 
```bash
datacontract export --format custom --template my_template.jinja datacontract.yaml
```

**Actual behavior :**
Use {{ _data_contract_ }} in your custom templates, but it exposes DCS fields only.

**Expected behaviour :**  
{{ _data_contract_ }} exposes all the fields relative to ODCS.

gautierc-thelio:
- [ ] Tests pass
- [ ] ruff format

Closes #814

### DIFF
--- a/datacontract/cli.py
+++ b/datacontract/cli.py
@@ -213,7 +213,6 @@ def export(
         typer.Option(help="[custom] The file path of Jinja template."),
     ] = None,
     spec: Annotated[
-        Optional[Spec],
         Spec,
         typer.Option(help="The format of the data contract to import. "),
     ] = Spec.datacontract_specification,
@@ -222,7 +221,9 @@ def export(
     Convert data contract to a specific format. Saves to file specified by `output` option if present, otherwise prints to stdout.
     """
 
-    # TODO exception handling
+    if spec is not None and format != ExportFormat.custom:
+        raise typer.BadParameter("--spec is only allowed with --format custom.")
+
     result = DataContract(
         data_contract_file=location,
         schema_location=schema,
--- a/datacontract/cli.py
+++ b/datacontract/cli.py
@@ -212,19 +212,30 @@ def export(
         Optional[Path],
         typer.Option(help="[custom] The file path of Jinja template."),
     ] = None,
+    spec: Annotated[
+        Optional[Spec],
+        Spec,
+        typer.Option(help="The format of the data contract to import. "),
+    ] = Spec.datacontract_specification,
 ):
     """
     Convert data contract to a specific format. Saves to file specified by `output` option if present, otherwise prints to stdout.
     """
+
     # TODO exception handling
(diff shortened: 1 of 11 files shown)


datacontract/data_contract.py
```diff
@@ -41,7 +41,7 @@ def __init__(
         self,
         data_contract_file: str = None,
         data_contract_str: str = None,
-        data_contract: DataContractSpecification = None,
+        data_contract: DataContractSpecification | OpenDataContractStandard = None,
```
jochenchrist, line 44: This change would require a switch in all instance methods, in particular in test() methods.

It feels correct that we should support ODCS spec in the `DataContract` class. However introducing this in would be a major enhancement  (and this PR is a bit specific).

jochenchrist: cc @simonharrer

datacontract/data_contract.py
```diff
@@ -250,8 +250,8 @@ def get_data_contract_specification(self) -> DataContractSpecification:
             inline_quality=self._inline_quality,
         )
 
-    def export(self, export_format: ExportFormat, model: str = "all", sql_server_type: str = "auto", **kwargs) -> str:
-        if export_format == ExportFormat.html or export_format == ExportFormat.mermaid:
+    def export(self, export_format: ExportFormat, model: str = "all", spec: Spec = Spec.datacontract_specification, sql_server_type: str = "auto", **kwargs) -> str:
+        if export_format == ExportFormat.html or export_format == ExportFormat.mermaid or spec == Spec.odcs:
```
simonharrer, line 254: We should enable the new resolve logic (v2) also for ExportFormat.custom. We do not need the spec as a parameter here.

datacontract/cli.py
```diff
@@ -212,19 +212,32 @@ def export(
         Optional[Path],
         typer.Option(help="[custom] The file path of Jinja template."),
     ] = None,
+    spec: Annotated[
```
simonharrer, line 215: We do not need the parameter, as we can automatically detect whether a YAML file is DCS or ODCS. Supporting a jinja template that can handle DCS and ODCS is not something we want to support. And it could be implemented with an instance of check anyway without this parameter.

simonharrer:
@gautierc-thelio thanks for your contribution! This is really important, and long overdue. We can simplify the implementation quite a bit, by staying close to how the HTML export itself is implemented. We do not need the `--spec` parameter, making the implementation much simpler. Would you be so kind to simplify your solution along those lines? I've added a few inline comments as pointers. Then, we can merge this quickly.

jochenchrist:
@gautierc-thelio

Can you incorporate the requested changes into this PR?

PR #842, opened 2025-07-31
feat: add extra Markdown export functionality for data contracts

julien-ch07:
The purpose here is to add every extra fields from Pydantic model in Markdown converter.

Pydantic models expose `model_extra` attribute that we can use to get every property that does not belong to model.

It displays accordingly to the data type (`dict`, `list`, `list[dict]` and so on) in the `Extra` part.

- [x] Tests pass
- [x] ruff format


tests/fixtures/markdown/export/expected.md
```diff
@@ -23,6 +23,24 @@ Max data processing per day: 10 TiB
 - **policies:** [{'url': 'https://example.com/privacy-policy', 'name': 'privacy-policy'}, {'description': 'External data is licensed under agreement 1234.', 'url': 'https://example.com/license/1234', 'name': 'license'}]
 - **billing:** 5000 USD per month
 - **noticePeriod:** P3M
+### Extra
```
jochenchrist: I would not call this "Extra".
Just add the additional objects (with headlines) and properties.

jochenchrist:
Thanks for your contribution!

PR #847, opened 2025-08-06
Fix: Support globbing for local JSON files

rahulwermunddmde:
Add unit tests using glob path to JSON files

- [x] Tests pass
- [x] ruff format


tests/fixtures/local-json-multiline/datacontract.yaml
```diff
@@ -0,0 +1,35 @@
+dataContractSpecification: 1.2.0
+id: my-data-contract-local
+info:
+  title: My Data Contract Local
+  version: 0.0.1
+  description: My Data Contract Description
+servers:
+  test:
+    type: local
+    path: ./fixtures/local-json-multiline/**/*.json
```
jochenchrist: Why "multiline"?

rahulwermunddmde: My main goal was to differentiate from the existing JSON fixtures. The data file contains multiple JSON lines, so "multiline". I now see that the `local-json` fixture also has a multiline JSON file. However, the corresponding test fails due to Soda quoting issues. We could replace that test with the new "multiline" one, if you think that's a good idea.

jochenchrist: We usually call these json files new line delimited, following the https://github.com/ndjson/ndjson-spec terminology.
These ND json files are pretty common when working with data sets and json.

rahulwermunddmde: Good point. That's the term I was actually looking for; it just didn't occur to me. I'll rename the fixture to `local-json-nd` if that's fine by you.

jochenchrist (approved):
LGTM

PR #869, opened 2025-08-22
feat: linting for all yaml datacontracts via yaml language server com…

benomahony:
This would be a huge quality life improvement for anyone editing a yaml data contract 

Screenshot 2025-08-22 at 12 29 34
…ment

- [x] Tests pass
- [x] ruff format

### DIFF
--- a/datacontract/schemas/datacontract-1.1.0.init.yaml
+++ b/datacontract/schemas/datacontract-1.1.0.init.yaml
@@ -1,4 +1,5 @@
-dataContractSpecification: 1.2.0
+# yaml-language-server: $schema=https://raw.githubusercontent.com/datacontract/datacontract-cli/main/datacontract/schemas/datacontract-1.1.0.schema.json
+dataContractSpecification: 1.1.0
 id: my-data-contract-id
 info:
   title: My Data Contract
@@ -9,7 +10,6 @@ info:
 #    name:
 #    url:
 #    email:
-
 
 ### servers
 
@@ -28,7 +28,6 @@ info:
 #  billing:
 #  noticePeriod:
 
-
 ### models
 
 # models:
@@ -39,7 +38,6 @@ info:
 #       my_field:
 #         type:
 #         description:
-
 
 ### definitions
 
@@ -53,7 +51,6 @@ info:
 #     example:
 #     pii:
 #     classification:
-
 
 ### servicelevels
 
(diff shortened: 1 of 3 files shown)


benomahony:
These tests fail because I added a comment to the schema so there is a diff but this is expected:

`tests/test_download_datacontract_file.py F.F.`

datacontract/schemas/datacontract-1.1.0.init.yaml
```diff
@@ -1,4 +1,5 @@
-dataContractSpecification: 1.2.0
+# yaml-language-server: $schema=https://raw.githubusercontent.com/datacontract/datacontract-cli/main/datacontract/schemas/datacontract-1.1.0.schema.json
```
jochenchrist, line 1: Please remove this init line, as it would become part of all data contracts.

datacontract/schemas/datacontract-1.2.0.init.yaml
```diff
@@ -1,3 +1,4 @@
+# yaml-language-server: $schema=https://raw.githubusercontent.com/datacontract/datacontract-cli/main/datacontract/schemas/datacontract-1.2.0.schema.json
```
jochenchrist, line 1: Please remove this init line, as it would become part of all data contracts.

datacontract/schemas/datacontract-1.2.0.init.yaml
```diff
@@ -89,3 +86,90 @@ info:
 #    cron: 0 0 * * 0
 #    recoveryTime: 24 hours
 #    recoveryPoint: 1 week
+dataContractSpecification: 1.2.0
```
jochenchrist, line 89: seems wrong

datacontract/schemas/datacontract-1.1.0.init.yaml
benomahony: That's kinda the whole point? This way all data contracts get linting, auto completion and type hints? (As long as the IDE has any yaml language server running)

jochenchrist: We rely on https://www.schemastore.org/ for auto completion. It works with naming conventions.

benomahony: If that works then let's just close this pr. I use neovim so would need some setup but my colleague using vscode didn't get any of the benefits until we installed a yaml language server. I didn't check if it worked for him without the schema comment

jochenchrist:
I am closing this PR, as we rely on https://www.schemastore.org/ for auto-completion.

PR #876, opened 2025-08-29
fix : odcs excel export Error filling properties 

dmaresma:
`WARNING:datacontract.export.excel_exporter:Error filling properties: sequence item 0: expected str instance, int found
Written result to excel.xlsx`

When the examples is an array of int the warning raise and stop the rendering

adding a map(str, property.examples) solve the ",".join() python statement

",".join(map(str, property.examples))

```yaml
...
- name: exclusionType
      busninessName: exclusionType
      logicalType: string
      physicalType: string
      required: true
      unique: false
      examples:
        - "subCategory"
    - name: exclusionCode
      busninessName: exclusionCode
      logicalType: integer
      physicalType: bigint
      required: true
      unique: false
      examples:
        - 624002
```

fix the property name `None` presentation with nested properties
before
image

after
image

- [ ] Tests pass
- [ ] ruff format

### DIFF
--- a/datacontract/export/excel_exporter.py
+++ b/datacontract/export/excel_exporter.py
@@ -283,7 +283,7 @@ def fill_single_property_template(
     sheet: Worksheet, row_index: int, prefix: str, property: SchemaProperty, header_map: dict
 ) -> int:
     """Fill a single property row using the template's column structure"""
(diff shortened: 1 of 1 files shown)


jochenchrist (approved):
LGTM

PR #881, opened 2025-09-04
fix rendering: schema instead of schema_ in odcs html

GuntherStruyf:
- [ ] Tests pass
- [ ] ruff format

### DIFF
--- a/datacontract/templates/datacontract_odcs.html
+++ b/datacontract/templates/datacontract_odcs.html
@@ -211,7 +211,7 @@
               </div>
             </div>
 
-            {% for schema in datacontract.schema_ %}
+            {% for schema in datacontract.schema %}
 
             <div class="mt-3 flow-root">
               <div class="-mx-4 -my-2 overflow-x-auto sm:-mx-6 lg:-mx-8">


datacontract/templates/datacontract_odcs.html
```diff
@@ -211,7 +211,7 @@ <h1 class="text-base font-semibold leading-6 text-gray-900">
               </div>
             </div>
 
-            {% for schema in datacontract.schema_ %}
+            {% for schema in datacontract.schema %}
```
jochenchrist, line 214: `schema_` is correct, as `schema` is a reserved word in Python.

GuntherStruyf: afaik `schema` is not a reserved word in python, neither in javascript; or do you mean reserved for other purposes in context of datacontract?

A validated odcs yaml contract will just have the field `datacontract.schema` and not `.schema_`
I ran into this when trying datacontract editor for odcs with adapted render.js:

        return nunjucks.render("datacontract_odcs.html", { datacontract: dataContract, formatted_date: local_date, 

When I change it to `schema`, the template is still properly generated and my odcs contract schemas _are_ rendered.

jochenchrist: >afaik schema is not a reserved word in python, 
You are correct. Pydantic is introducing a `schema` property (`BaseModel.schema`). So, workaround is to use `schema_` is we have a "schema" property.

PR #902, opened 2025-09-29
901 implement decimal support for databricks and spark testing + fix test spark startup

Tyranwyn:
- [ ] Tests pass
- [ ] ruff format

### DIFF
--- a/tests/test_test_dataframe.py
+++ b/tests/test_test_dataframe.py
@@ -1,4 +1,5 @@
 from datetime import datetime
+from decimal import Decimal
 
 import pytest
 from dotenv import load_dotenv
@@ -9,7 +10,7 @@ from pyspark.sql.types import (
     StringType,
     StructField,
     StructType,
-    TimestampType,
+    TimestampType, DecimalType,
 )
 
 from datacontract.data_contract import DataContract
@@ -22,7 +23,7 @@ load_dotenv(override=True)
 
 
 @pytest.fixture(scope="session")
-def spark(tmp_path_factory) -> SparkSession:
+def spark(tmp_path_factory, request) -> SparkSession:
     """Create and configure a Spark session."""
     spark = (
         SparkSession.builder.appName("datacontract-dataframe-unittest")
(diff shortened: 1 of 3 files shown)


tests/test_test_dataframe.py
```diff
@@ -35,10 +37,13 @@ def spark(tmp_path_factory) -> SparkSession:
             "spark.jars.packages",
             "org.apache.spark:spark-sql-kafka-0-10_2.12:3.5.5,org.apache.spark:spark-avro_2.12:3.5.5",
         )
+        .config('spark.driver.host', '127.0.0.1')
```
0xpugsley, line 41: What i am really missing in that fixuture is
```python
SparkSession.builder.appName("datacontract-dataframe-unittest") 
.master("local[*]") 
# and
.config("spark.ui.enabled", false)
```

first tells spark that there is no cluster, second to not bother with ui server because it is unit test.

tests/test_test_dataframe.py
```diff
@@ -35,10 +37,13 @@ def spark(tmp_path_factory) -> SparkSession:
...
         .getOrCreate()
     )
     spark.sparkContext.setLogLevel("WARN")
     print(f"Using PySpark version {spark.version}")
+
```
0xpugsley, line 48: AFAIK this can be done a bit differently, no need to take request explicitly
```python
  yield spark
  spark.stop()
```
but with session scope and local master idk if that will have much difference, spark should be stopped with test process

tests/test_test_dataframe.py
0xpugsley: BTW I forgot—with master there is no need for setting driver host

Tyranwyn: when removing spark.driver.host i get following error
image

0xpugsley: image

You can check it like (comment out the setter before)
```python
print(spark.sparkContext.getConf().get("spark.driver.host"))
```

It is not wrong to set (probably) but it is unusual and better not to do it if not necessary. It can be something with your network or JVM setup.

I have checked that test (main branch) and it works, `spark.driver.host` points to my hostname

tests/test_test_dataframe.py
```diff
@@ -22,7 +25,7 @@
 
 
 @pytest.fixture(scope="session")
-def spark(tmp_path_factory) -> SparkSession:
+def spark(tmp_path_factory, request) -> Generator[SparkSession, Any, None]:
```
0xpugsley: request is unneeded if generator is used

Issue #909, opened 2025-10-03
Issue parsing tables starting with a number for BigQuery Server


PeteZ238:
BigQuery supports [flexible column names](https://cloud.google.com/bigquery/docs/schemas#flexible-column-names) and [table names](https://cloud.google.com/bigquery/docs/tables#table_naming). However, datacontract-cli appears to have some issues parsing tables starting with a number correctly.

If the table name starts with an underscore or a letter, the tests run with no issues. This is both at column and table level.

```yaml
schema:
  - name: 501_analytics_table
    logicalType: object
    physicalType: table
    physicalName: _501_analytics_table
```

<img width="1117" height="541" alt="Image" src="https://github.com/user-attachments/assets/800fe7f4-ef59-45b7-81b3-5a70055984cd" />

If the table starts with a number and is defined in the contract (ODCS yaml) unquoted:

```yaml
schema:
  - name: 501_analytics_table
    logicalType: object
    physicalType: table
    physicalName: 501_analytics_table
```

Then the `required` and `unique` tests will not run. Furthermore, custom SQL quality tests for the table will run but not for specific columns.

```bash
🟠 data contract has warnings. Found the following warnings:
1) 501_analytics_table.COUNTRY_DESC Check that field COUNTRY_DESC has no missing values: None
2) 501_analytics_table.COUNTRY_DESC Custom sql test for null values in a column: None
3) 501_analytics_table.SENSOR_TYPE Check that field SENSOR_TYPE has no missing values: None
4) 501_analytics_table.SENSOR_HOURLY_TIMESTAMP Check that field SENSOR_HOURLY_TIMESTAMP has no missing values: None
5) 501_analytics_table.SENSORS_PER_MARKET Check that field SENSORS_PER_MARKET has no missing values: None
6) 501_analytics_table.MINIMUM_SENSOR_VALUE Check that field MINIMUM_SENSOR_VALUE has no missing values: None
7) 501_analytics_table.MAXIMUM_SENSOR_VALUE Check that field MAXIMUM_SENSOR_VALUE has no missing values: None
8) 501_analytics_table.AVERAGE_SENSOR_VALUE Check that field AVERAGE_SENSOR_VALUE has no missing values: None
```

If the table starts with a number and it is defined in the contract quoted, all tests fail to run.

```yaml
schema:
  - name: 501_analytics_table
    logicalType: object
    physicalType: table
    physicalName: "`501_analytics_table'"
```

```bash
[...]

PR #910, opened 2025-10-03
feat(BIG QUERY): Add support for BQ flexible schema in Data Contract Checks

PeteZ238:
Resolves issue #909 

- [x] Tests pass - sort of, see below
- [x] ruff format
- [x] README.md updated (if relevant) - not required

Unable to run all unit tests locally as they include integration tests I'm no able to run. I was hoping to run and validate everything in the CI but I don't have permissions to trigger it.

image

The unit tests for `data_contract_checks.py` ran with no issues (`tests/test_data_contract_checks.py`):

image

Testing the specific functionality I amended / updated works as intended with a known "problematic" ODCS data contract and a known working one.

image

### DIFF
--- a/datacontract/engines/data_contract_checks.py
+++ b/datacontract/engines/data_contract_checks.py
@@ -15,7 +15,7 @@ from datacontract.model.run import Check
 class QuotingConfig:
     quote_field_name: bool = False
     quote_model_name: bool = False
-    quote_with_backticks_model_name: bool = False
+    quote_model_name_with_backticks: bool = False
 
 
 def create_checks(data_contract_spec: DataContractSpecification, server: Server) -> List[Check]:
@@ -36,11 +36,13 @@ def to_model_checks(model_key, model_value, server: Server) -> List[Check]:
 
     check_types = is_check_types(server)
 
-    quoting_config = QuotingConfig(
-        quote_field_name=server_type in ["postgres", "sqlserver"],
-        quote_model_name=server_type in ["postgres", "sqlserver"],
-        quote_with_backticks_model_name=server_type == "bigquery",
+    type1 = server.type if server and server.type else None
+    config = QuotingConfig(
+        quote_field_name=type1 in ["postgres", "sqlserver"],
+        quote_model_name=type1 in ["postgres", "sqlserver"],
+        quote_model_name_with_backticks=type1 == "bigquery",
     )
+    quoting_config = config
 
     for field_name, field in fields.items():
         checks.append(check_field_is_present(model_name, field_name, quoting_config))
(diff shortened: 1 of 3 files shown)


.gitignore
```diff
@@ -330,3 +330,6 @@ fabric.properties
 # Generated test files
 **/.tmp
 /.claude/settings.local.json
+
```
jochenchrist: please clean this again

datacontract/engines/data_contract_checks.py
```diff
@@ -38,12 +39,13 @@ def to_model_checks(model_key, model_value, server: Server) -> List[Check]:
     quoting_config = QuotingConfig(
         quote_field_name=server_type in ["postgres", "sqlserver"],
         quote_model_name=server_type in ["postgres", "sqlserver"],
+        bq_flexible_model_name=server_type == "bigquery",
```
jochenchrist: please change `bq_flexible_model_name` to `quote_with_backticks`

datacontract/engines/data_contract_checks.py
```diff
@@ -82,9 +84,11 @@ def to_model_checks(model_key, model_value, server: Server) -> List[Check]:
     return checks
 
 
-def checks_for(model_name, quote_model_name: bool):
-    if quote_model_name:
+def checks_for(model_name: str, quoting_config: QuotingConfig, check_type: str) -> str:
+    if quoting_config.quote_model_name:
         return f'checks for "{model_name}"'
+    elif quoting_config.bq_flexible_model_name and check_type not in ["field_is_present", "field_type"]:
```
jochenchrist: why is this only needed for these check_types?
Can the backticke be applied to any bq fields?

.gitignore
PeteZ238: Done :)

datacontract/engines/data_contract_checks.py
PeteZ238: Renamed to `quote_with_backticks_model_name` as it only affects the model name and not the field names

datacontract/engines/data_contract_checks.py
PeteZ238: To the best of my understanding, this is either a BQ or Soda idiosyncrasy. However, I am not very familiar with the codebase so I could very well be wrong.

Essentially, it seems tests where we query the metadata of a table and not the actual table run fine without the backticks regardless of whether the table name starts with a number or a letter, etc. I made a note of this in the issue.

However, if  we do add the backticks, the metadata tests will fail to run (again, regardless of the table name).

image

Hence why I made the distinction in what tests to not apply the backticks.

jochenchrist:
OK, LGTM. Thank you for your contribution!

PR #923, opened 2025-10-31
Fix #922: validate the contract against the custom schema before converting

### LINKED ISSUE #922: Bug: Custom schema validation is skipped for ODCS format contracts
### Description

When using the `--schema` option with a custom JSON Schema to validate a data contract in ODCS (Open Data Contract Standard) format, the validation is silently **skipped**. The CLI reports the contract as valid even when it violates constraints defined in the custom schema (e.g., enum values).

### Steps to Reproduce

1. Create an ODCS contract with an invalid enum value:

```yaml
# contract-to-validate.yaml
version: "1.0.0"
kind: DataContract
apiVersion: v3.0.1
id: "test-123"
status: active
domain: "INVALID_DOMAIN"  # This should fail validation if we use the custom schema below
```

2. Create a custom schema with enum constraint:

```json
{
  "$schema": "https://json-schema.org/draft/2019-09/schema",
  "type": "object",
  "properties": {
    "domain": {
      "type": "string",
      "enum": ["VALID_DOMAIN"]
    }
  },
  "required": ["version", "apiVersion", "kind", "id", "status"]
}
```

3. Run validation

```bash
datacontract lint contract-to-validate.yaml --schema custom-schema.json
```

Expected Behavior
The validation should **fail** with an error message:

```bash
1) Check that data contract YAML is valid: data.domain must be one of ['VALID_DOMAIN']
````
Actual Behavior
The validation **passes** incorrectly:
```bash
🟢 data contract is valid. Run 1 checks. Took 0.001373 seconds.
```

Root Cause
In `datacontract/lint/resolve.py` the `_resolve_data_contract_from_str` function converts ODCS contracts before validating them against the custom schema:

```python
def _resolve_data_contract_from_str(
    data_contract_str, schema_location: str = None, inline_definitions: bool = False, inline_quality: bool = False
) -> DataContractSpecification:
    yaml_dict = _to_yaml(data_contract_str)


    if is_open_data_contract_standard(yaml_dict):
        logging.info("Importing ODCS v3")
        # if ODCS, then validate the ODCS schema and import to DataContractSpecification directly
        odcs = parse_odcs_v3_from_str(data_contract_str)

        data_contract_specification = DataContractSpecification(dataContractSpecification="1.2.1")
        return import_from_odcs(data_contract_specification, odcs)
...
```

When `is_open_data_contract_standard()` returns True, the function:

1. Converts the ODCS contract to Data Contract Specification format
2. Returns immediately
3. Never calls `_validate_data_contract_specification_schema()]`

Proposed Solution
Validate the contract against the custom schema and call `_validate_data_contract_specification_schema` before converting from ODCS format

[Proposed Fix](https://github.com/datacontract/datacontract-cli/pull/923)

YLibert:
This fixes the issue [#922](https://github.com/datacontract/datacontract-cli/issues/922)
- [x] Tests pass
- [x] ruff format

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,11 +9,12 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 
-- Support for nested arrays in odcs v3 importer
+- import: Support for nested arrays in odcs v3 importer
+- lint: ODCS schema is now checked before converting
 
 ### Fixed
 
-- Excel exporter now exports critical data element
+- export: Excel exporter now exports critical data element
 
 
 ## [0.10.36] - 2025-10-17
--- a/datacontract/lint/resolve.py
+++ b/datacontract/lint/resolve.py
@@ -1,6 +1,8 @@
+import importlib.resources as resources
 import logging
 import os
 import warnings
+from pathlib import Path
 
 import fastjsonschema
 import yaml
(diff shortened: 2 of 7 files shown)


jochenchrist:
Thank you for your PR.
I also added a unit test and the changelog.

PR #941, opened 2025-11-11
DuckDB DDL generation produces invalid empty STRUCT() for objects without defined properties

### LINKED ISSUE #940: DuckDB DDL generation produces invalid empty STRUCT() for objects without defined properties
### Background:
There are legitimate scenarios where a field is defined as type: object without explicitly defining its properties. An example is for adding custom attributes:

`      - name: "properties"
        logicalType: "object"
        physicalType: "object"
        description: "Additional custom attributes."`

================================================================

ISSUE:
When generating DuckDB DDL from JSON Schema definitions, objects that have "type": "object" but no properties field are incorrectly converted to empty STRUCT() definitions, which is invalid DuckDB syntax.

This issue occurs when running `datacontract test` with a data contract containing objects without defined properties.

**Error:**
Parser Error: syntax error at or near ")"

didipayson:
- [x] Tests pass
- [x] ruff format

Fixes #940

Summary:
This PR fixes a bug where DuckDB DDL generation creates invalid empty STRUCT() syntax for objects that have type: object but no defined properties.

Changes:

- Modified DDL generation logic to use JSON type instead of STRUCT() for objects without properties
- Added test case to verify the fix

Why JSON type?
Objects without defined properties are flexible (e.g., metadata fields, custom attributes), making DuckDB's JSON type the correct representation rather than attempting to create an empty STRUCT().


datacontract/export/duckdb_type_converter.py
```diff
@@ -48,6 +48,9 @@ def convert_to_duckdb_json_type(field: Field) -> None | str:
     if datacontract_type.lower() in ["array"]:
         return convert_to_duckdb_json_type(field.items) + "[]"  # type: ignore
     if datacontract_type.lower() in ["object", "record", "struct"]:
+        # If no fields are defined, treat as generic JSON
+        if field.fields is None or len(field.fields) == 0:
+            return "JSON"
```
jochenchrist, line 53: @simonharrer what do you think? How should an object without fields be treated?

simonharrer: Three options
- introduce a general fallback to `text`
- use `MAP(STRING, STRING)` for this particular case
- use `JSON` for this particular case (requires JSON extension, but this is autoloaded)

I think they are all fine. The solution will probably always be to use the `physicalType` to override this. JSON is more convenient, and the fallback to text will lead the user more quickly to enter the right physicalType.

So no really wrong choice here. JSON seems fine.

PR #952, opened 2025-11-22
added support for map logical type since it was missing

robert-altmiller:
The map logical type was missing for fields which have a map datatype.  The physical type was there.  Without the logic type added the Map<key, val> was not written correctly in the data contract.

Testing was done with map field which looks like the following:
image

This is the fix we implemented:
image

- [ ] Tests pass
- [ ] ruff format


robert-altmiller:
Variant is no longer a valid logical type based on linting so I had to make another change:
image

PR #956, opened 2025-11-30
Export and Import Excel to (ODCS v3) YAML Bug Fixes

christianvanr:
This pull request addresses the following issues, when using the excel importer/exporter:

**Import**: Fixed an issue where YAML output becomes corrupted when importing quality rules from Excel files (ODCS v3), specifically when using the mustBe and mustBeNot expectation fields. 

_The error message was: mustBe: !!python/object/apply:decimal.Decimal_

Steps to reproduce without the fix:

datacontract import --format excel --source .xlsx --output <datacontract_name>.yaml
datacontract lint <datacontract_name>.yaml

**Export**: The mustBeGreaterOrEqualTo and mustBeLessOrEqualTo fields were not implemented, even though they are available as dropdown options in the Excel file. Note: these are not the "...Than..." variants.

The shipments.xlsx file has been updated to include test cases for the corrupted YAML scenario.

**Further:**
I implemented this fix in a way that I believe has minimal impact, as I am not familiar with the background of the code base. So the fixes are best effort, maybe i did something totally unnecessary

For now, I think these solutions are sufficient. In the future, I plan to make the implementation more robust when i have more available time. 
Additionally, not sure, but i think string results (or boolean) should also be supported for mustBe and mustBeNot. (with correct linting etc) and the ranges should support integers as well as floats


jochenchrist:
Good catch, thank you!

PR #964, opened 2025-12-03
Allow the specification of csv schema parameters in server object instead of simple inferring it using duckdb

liepelt-cornelius:
- [x] Tests pass (some kafka tests and the custom exporter fail but are unrelated to my change)
- [x] ruff format

As it is my first time contributing here please point out what is missing or needs to be fixed.


jschoedl:
Hi @liepelt-cornelius, thank you for your PR!
Unfortunately, this has been lying around here for quite some time now... `duckdb_connection.py` has changed too much for these changes to fit anymore. But the feature itself would be valuable and we would appreciate an updated PR on this!

PR #978, opened 2025-12-15
feat: export custom model (like as dbt models .sql)

AntoineGiraud:
## what 

Hi, this is a PR to add the `--schema_name` when we do a custom export

example :
`datacontract export tests/fixtures/custom/export/datacontract.yaml --format custom --template tests/fixtures/custom/export_model/template.sql --schema-name line_items`

## why

i wasn't able to do it previously and it felt odd to have a for loop to [export a custom dbt model](https://github.com/datacontract/datacontract-cli?tab=readme-ov-file#customized-dbt-model) ... that is supposed to be 1 model per file

## to be discused

I tried to update the README.md acordingly

This PR is in draft to open the discusion with materials to chat on

i'm also puzzled on if the "data_contract" passed to the `template.file` will be an ODCS exclusively
or could it still be the ol' datacontract spec ?!

though for now the method is
`def to_custom(data_contract: OpenDataContractStandard, template_path: Path) -> str:`

## PushToProd checks

- [x] Tests pass
- [x] ruff format

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Added `ci` command for CI/CD-optimized test runs: multi-file support, GitHub Actions annotations and step summary, Azure DevOps annotations, `--fail-on` flag, `--json` output
 - Added `changelog` command and API endpoint (#1118)
 - Added opt-in `--all-errors` mode for `datacontract lint` to report all JSON Schema validation errors, with matching `all_errors` support in the Python library and API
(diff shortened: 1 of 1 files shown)


jschoedl:
Hi @AntoineGiraud, thank you for your contribution! Sorry, this was lying around here for quite some time now.

I don't think we need such extensive testing for this feature, so I just went ahead and reduced them a bit. Otherwise LGTM, thanks!

AntoineGiraud:
Toll ! vielen dank

FYI: the 2 table test was juste to be sure the 2nd model was exported ;)

PR #982, opened 2025-12-16
fix(export-pydantic): generate decimal.Decimal type for decimals

yaghmaie:
Currently, the pydantic model exporter generates `float` for the fields marked with `type: decimal`.

```
some_field:
  type: decimal
```
converts into:
```
class SomeClass(pydantic.BaseModel):
    f: typing.Optional[float]
```
instead of:
```
class SomeClass(pydantic.BaseModel):
    f: typing.Optional[decimal.Decimal]
```

This is important to keep the types consistent across different formats.
For instance, the data that's dumped from a generated pydantic model should be encoded by a generated avro schema with no trouble, while currently avro encoding fails because `float`s are provided instead of explicit `Decimal`s.
This PR fixes such cases.

- [x] Tests pass - 1 added to `tests/test_export_pydantic.py` and all of them in the file pass
- [x] ruff format - Only applied for the modified files in the PR 
- [x] README.md updated (if relevant) - Not needed
- [x] CHANGELOG.md entry added - Under unreleased fix


jochenchrist (approved):
LGTM

PR #986, opened 2025-12-19
Feature: Add Teradata SQL importer

rory-data:
This pull request adds support for importing Teradata SQL DDL files into ODCS data contracts, while also improving typing, fixing linter issues, and simplifying code in the SQL importer module.

#### Key Changes:

- Added Teradata dialect support with mapping for Teradata-specific types (BYTEINT, BYTE, VARBYTE, INTERVAL types)
- Improved type annotations throughout `sql_importer.py` and `odcs_helper.py` using modern Python typing syntax (e.g. `dict[str, str]` instead of `Dict[str, str]`)
- Refactored code for better maintainability using match statements and simplified conditionals

- [x] Tests pass
- [x] ruff format

### DIFF
--- a/datacontract/imports/sql_importer.py
+++ b/datacontract/imports/sql_importer.py
@@ -91,6 +91,7 @@ def import_sql(import_format: str, source: str, import_args: dict[str, str] | No
 
             col_type = to_col_type(column, dialect)
             is_primary_key = get_primary_key(column)
+            precision, scale = get_precision_scale(column)
 
             prop = create_property(
                 name=column.this.name,
@@ -98,8 +99,8 @@ def import_sql(import_format: str, source: str, import_args: dict[str, str] | No
                 physical_type=col_type,
                 description=get_description(column),
                 max_length=get_max_length(column),
-                precision=get_precision_scale(column)[0],
-                scale=get_precision_scale(column)[1],
+                precision=precision,
+                scale=scale,
                 primary_key=is_primary_key,
                 primary_key_position=primary_key_position if is_primary_key else None,
                 required=column.find(sqlglot.exp.NotNullColumnConstraint) is not None,
@@ -204,6 +205,8 @@ def to_col_type_normalized(column: ColumnDef) -> str | None:
     Returns:
         The normalized SQL type string or None if not found.
     """
+    if column.args["kind"] is None:
+        return None
     col_type = column.args["kind"].this.name
     return col_type.lower() if col_type is not None else None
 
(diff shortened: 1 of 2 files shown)


datacontract/imports/odcs_helper.py
```diff
@@ -1,6 +1,8 @@
 """Helper functions for creating ODCS (OpenDataContractStandard) objects."""
 
-from typing import Any, Dict, List
+from __future__ import annotations
```
jochenchrist, line 3: Why?

datacontract/imports/sql_importer.py
```diff
@@ -1,9 +1,17 @@
+"""SQL importer for data contracts.
```
jochenchrist, line 1: This doc does not help

datacontract/imports/odcs_helper.py
rory-data: The `annotations` import was added to resolve test failures due to the introduction of  the `|` option type hint. Have just tested again with that commented out to reproduce the original test error, but it looks like a subsequent commit has removed the need for it altogether as the tests are passing.

datacontract/imports/sql_importer.py
rory-data: The callout on the generated docstrings quality is fair; a good reminder to not let a bad habit form there.

tests/test_import_sql_teradata.py
```diff
@@ -0,0 +1,236 @@
...
+    expected = """
+apiVersion: v3.1.0
+kind: DataContract
+id: my-data-contract
+name: My Data Contract
+version: 1.0.0
+status: draft
+servers:
+  - server: teradata
+    type: teradata
```
simonharrer, line 45: The type teradata is not part of the supported types in ODCS (see https://bitol-io.github.io/open-data-contract-standard/latest/infrastructure-servers/). We would need to add that to the standard. Until then, you have to use custom, and use a custom property to specify the serverType:

```
server: teradata
type: custom
customProperties:
- property: customType
  value: teradata
```

davidb-tada:
@rory-data I am interested in introducing Teradata support. May I help with this PR ?

jschoedl:
Hi @rory-data, thank you for your contribution!

As discussed, I'm closing this PR now, but we would welcome a more focused change (just adding the Teradata dialect and type mappings to the SQL importer).

@davidb-tada We would appreciate your work on this! Maybe this PR can still serve as a reference for the mappings.

PR #1017, opened 2026-01-14
fix BigQuery import bug when the field is repeated

### LINKED ISSUE #737: Fix Bug:  Differentiate between the Nested and Repeated fields for bigquery when executing datacontract import.

amir-qasemi:
The repeated mode in BigQuery is not handled. This merge request fixes this problem.
This merge request fixes #737 

- [x] Tests pass
- [x] ruff format

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Fixed
 
 - Fix BigQuery import for repeated fields 
+- Make Markdown export compatible with XHTML by replacing `<br>` with `<br />` (#1030)
 
 ## [0.11.4] - 2026-01-19
 
--- a/datacontract/engines/data_contract_checks.py
+++ b/datacontract/engines/data_contract_checks.py
@@ -72,7 +72,7 @@ def to_schema_checks(schema_object: SchemaObject, server: Server) -> List[Check]
 
     type1 = server.type if server and server.type else None
     config = QuotingConfig(
-        quote_field_name=type1 in ["postgres", "sqlserver"],
+        quote_field_name=type1 in ["postgres", "sqlserver", "azure"],
         quote_model_name=type1 in ["postgres", "sqlserver"],
         quote_model_name_with_backticks=type1 == "bigquery",
     )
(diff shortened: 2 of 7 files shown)


amir-qasemi:
Moved the change log to the unreleased section

jochenchrist (approved):
LGTM

PR #1044, opened 2026-02-06
Use relationship label if present in custom properties

garrettsutula:
When `relationships[].customProperties['property=label']` is specified, that value will be used as the label in the relationship line drawn between schemas in the exported Mermaid diagram.

This makes it possible for data model authors to further annotate relationships in a way that will generate meaningful diagrams following a UML-like approach for roles described here: https://stackoverflow.com/questions/16732607/role-name-in-association-relationship

Generally, this gives the user a way to override the default display behavior if desired for whatever reason which I don't think merits specification in the standard because it is not really a primary concern of the data contract.

- [X] Tests pass
- [X] ruff format

I **did** modify `tests/fixtures/export/datacontract.odcs.yaml` and `tests/test_export_mermaid.py` to add a relationship test assertion and run locally but adding an entity to `datacontract.odcs.yaml` caused 7 other test cases to fail so I did not include that in the scope of this work.

### DIFF
--- a/datacontract/catalog/catalog.py
+++ b/datacontract/catalog/catalog.py
@@ -32,7 +32,11 @@ def create_data_contract_html(contracts, file: Path, path: Path, schema: str):
     odcs = data_contract.get_data_contract()
     file_without_suffix = file.with_suffix(".html")
     html_filepath = path / file_without_suffix
-    html_filepath.parent.mkdir(parents=True, exist_ok=True)
+    try:
+        html_filepath.parent.mkdir(parents=True, exist_ok=True)
+    except FileExistsError:
+        if not html_filepath.parent.is_dir():
+            raise
     with open(html_filepath, "w", encoding="utf-8") as f:
         f.write(html)
     contracts.append(
--- a/datacontract/cli.py
+++ b/datacontract/cli.py
@@ -431,7 +431,11 @@ def catalog(
     enable_debug_logging(debug)
 
     path = Path(output)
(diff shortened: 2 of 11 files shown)


datacontract/export/mermaid_exporter.py
```diff
@@ -37,13 +45,15 @@ def to_mermaid(data_contract: OpenDataContractStandard) -> str | None:
                     if prop.relationships:
                         for rel in prop.relationships:
                             ref_target = getattr(rel, 'to', None) or getattr(rel, 'ref', None)
+                            custom_label = _get_custom_property_value(rel, "label")
```
jochenchrist, line 48: You are calling the method with a wrong parameter type (Relationship it not a SchemaProperty)

garrettsutula: Whoops, pushed a correction. I didn't have my IDE fully initialized and forgot the runtime doesn't check 'em (TS & C# mostly for me).

jschoedl:
Hi @garrettsutula, thank you for the PR! Sorry, this was lying around here for quite some time now.

Is this customization something that is crucial for your usage of the Mermaid export? I guess it would be also possible using a custom post-processing script or similar?

There are likely many more things one could customize in the Mermaid diagrams. In general, this is not the priority for the datacontract-cli. Therefore I'd lean towards closing this as out of scope.

garrettsutula:
No, not critical I'll close this. Fairly straightforward to use a custom template instead.

PR #1069, opened 2026-02-22
add 'dqx' as test engine for DQT tests within datacontract

gkoenig:
Add an alternative test engine to Soda, by introducing Databricks' DQX framework ... which obviously requires _Databricks_ as server_type. Since DQX is able to read datacontracts in ODCS format natively, no conversion is required.
Initial extension offers DQX as test_engine in programmatic mode.

- [+] Tests pass
- [+] ruff format
- [+] README.md updated (if relevant)
- [-] CHANGELOG.md entry added

### DIFF
--- a/README.md
+++ b/README.md
@@ -142,6 +142,31 @@ run = data_contract.test()
 if not run.has_passed():
     print("Data quality validation failed.")
     # Abort pipeline, alert, or take corrective actions...
+
+# run quality checks with DQX engine (Databricks server type only)
+# requires: pip install datacontract-cli[dqx]
+data_contract_dqx = DataContract(
+  data_contract_file="odcs.yaml",
+  server="production",
+  test_engine="dqx",
+)
+run_dqx = data_contract_dqx.test()
+
+# access all executed DQX rule checks
+all_checks = run_dqx.checks
+
+# access failed/error/warning DQX rule checks
+failed_checks = [c for c in run_dqx.checks if c.result in ("failed", "error", "warning")]
+
+for check in failed_checks:
+  print(f"[{check.result}] {check.name} ({check.model})")
+  print(f"reason: {check.reason}")
+
+# full structured output
+print(run_dqx.pretty())
+
+if not run_dqx.has_passed():
+  print("DQX data quality validation failed.")
 ```
 
 ## How to
@@ -236,6 +261,7 @@ A list of available extras:
 | Avro Support            | `pip install datacontract-cli[avro]`       |
 | Google BigQuery         | `pip install datacontract-cli[bigquery]`   |
 | Databricks Integration  | `pip install datacontract-cli[databricks]` |
+| DQX (Databricks quality checks) | `pip install datacontract-cli[dqx]` |
 | DuckDB (local/S3/GCS/Azure file testing) | `pip install datacontract-cli[duckdb]` |
 | Iceberg                 | `pip install datacontract-cli[iceberg]`    |
 | Kafka Integration       | `pip install datacontract-cli[kafka]`      |
--- a/datacontract/api.py
+++ b/datacontract/api.py
@@ -323,11 +323,18 @@ async def test(
             examples=["https://api.datamesh-manager.com/api/test-results"],
         ),
     ] = None,
+    test_engine: Annotated[
+      str,
+      Query(
+        description="The engine used for quality checks. Supported values: soda (default), dqx (Databricks only).",
+        examples=["soda", "dqx"],
+      ),
+    ] = "soda",
 ) -> Run:
     check_api_key(api_key)
     logging.info("Testing data contract...")
     logging.info(body)
-    return DataContract(data_contract_str=body, server=server, publish_url=publish_url).test()
+    return DataContract(data_contract_str=body, server=server, publish_url=publish_url, test_engine=test_engine).test()
 
 
 @app.post(
--- a/datacontract/cli.py
+++ b/datacontract/cli.py
@@ -161,6 +161,12 @@ def test(
         typer.Option(help="SSL verification when publishing the data contract."),
     ] = True,
     debug: debug_option = None,
+    test_engine: Annotated[
+        str,
+        typer.Option(
+            help="The engine used for quality checks. Supported values: `soda` (default), `dqx` (Databricks only)."
+        ),
+    ] = "soda",
 ):
     """
     Run schema and quality tests on configured servers.
@@ -177,6 +183,7 @@ def test(
         publish_url=publish,
         server=server,
         ssl_verification=ssl_verification,
+        test_engine=test_engine,
     ).test()
     if logs:
         _print_logs(run)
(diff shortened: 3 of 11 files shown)


jochenchrist:
We need to make a strategic decision, if we want to support datasource-specific engines.
I like the idea, but it may cause inconsistent behavior for checks across different data sources,

dmaresma:
> We need to make a strategic decision, if we want to support datasource-specific engines. I like the idea, but it may cause inconsistent behavior for checks across different data sources,

The performance of native data quality engines should primarily be supported and optimized by the data platform providers themselves. From my personal experience, I have encountered limitations and issues with Snowflake DMFs. Having the ability to choose alternative engines is therefore important—not only from a technical standpoint, but also by principle, as it aligns with anti‑trust and vendor‑neutrality considerations.
SODA is a solid option, and DQT is also a viable alternative. What matters most is preserving optionality.
ODCS maintains the right level of abstraction, while the CLI gives users the flexibility to decide how they want to integrate and execute data quality checks. This separation is key: it allows innovation and specialization at the engine level without locking users into a single implementation.
Regarding the concern about inconsistent behavior across data sources if we support datasource‑specific engines: I see this as a manageable trade‑off rather than a blocker. In practice, we already deal with feature gaps and behavioral differences between platforms. Clear contracts, documented semantics, and conformance tests can mitigate most of these inconsistencies, while still allowing us to benefit from native optimizations where they make sense.
As a side note, I will be migrating my Snowflake workloads to Databricks in the near future, and data contracts will play an important role in securing and de‑risking that transition. This is a concrete example of why abstraction and portability matter—we already manage platform-specific feature defects today, and data contracts help turn those differences into explicit, controlled decisions rather than hidden risks.

gkoenig:
> We need to make a strategic decision, if we want to support datasource-specific engines. I like the idea, but it may cause inconsistent behavior for checks across different data sources,

Thanks for your feedback Jochen. 
Fully understand thorough thinking about major changes.....looking forward to you decision.

pocelka:
@gkoenig Although I'm not involved with the team maintaining CLI, I find your PR quite useful. Of course depending on what @jochenchrist will decide.

However I think that there is a small problem in the code or the approach you took. You probably need to exclude specific dataset level checks which user might configure in the contract. For example foreign key check as it depends on data from another DataFrame. And potentially any validation check that is not from the standard set of the checks; for example how would you deal with custom check that was [registered](https://databrickslabs.github.io/dqx/docs/reference/quality_checks/#using-python-function) with DQX? As an alternative the limitations should be described in readme.md.

Overall, this might be super useful for the people who use DQX.

jochenchrist:
We are considering a dqx sync command to convert and sync ODCS to a dqx project, so that dqx would run these tests directly.
```
datacontract dqx sync
```

Of course this should include quality checks with custom dqx tests

What do you think?

pocelka:
Hi, can you elaborate little bit more on how the sync command would work?

gkoenig:
> @gkoenig Although I'm not involved with the team maintaining CLI, I find your PR quite useful. Of course depending on what @jochenchrist will decide.
> 
> However I think that there is a small problem in the code or the approach you took. You probably need to exclude specific dataset level checks which user might configure in the contract. For example foreign key check as it depends on data from another DataFrame. And potentially any validation check that is not from the standard set of the checks; for example how would you deal with custom check that was [registered](https://databrickslabs.github.io/dqx/docs/reference/quality_checks/#using-python-function) with DQX? As an alternative the limitations should be described in readme.md.
> 
> Overall, this might be super useful for the people who use DQX.

thanks for your feedback @pocelka . You are absolutely right, there will be some fixes/extensions/etc to be worked on....as the current state of the PR was more a "kickstart" to figure out if it will make it to _main_ ;)

gkoenig:
> We are considering a dqx sync command to convert and sync ODCS to a dqx project, so that dqx would run these tests directly.
> 
> ```
> datacontract dqx sync
> ```
> 
> Of course this should include quality checks with custom dqx tests
> 
> What do you think?

thanks for your reply @jochenchrist 

What do you mean by "sync ODCS to a dqx project" ? ...creating an isolated small python project which gets the ODCS contract and uses the  dqx library to run the tests?

jochenchrist:
> What do you mean by "sync ODCS to a dqx project" ? ...creating an isolated small python project which gets the ODCS contract and uses the dqx library to run the tests?

Yes, idea is to convert ODCS to dbx tests and let dbx execute the tests. Should also work with existing databricks projects.

How do your Databricks data product code structure look like?

gkoenig:
> > What do you mean by "sync ODCS to a dqx project" ? ...creating an isolated small python project which gets the ODCS contract and uses the dqx library to run the tests?
> 
> Yes, idea is to convert ODCS to dbx tests and let dbx execute the tests. Should also work with existing databricks projects.
> 
> How do your Databricks data product code structure look like?

ok, got it, thanks @jochenchrist 
Actually, in one project we use datacontract-cli to validate the datacontract "schema" against UnityCatalog and use dqx library to run dqt checks, which are defined in dedicated yaml files (since we also want to cover dqt checks for data-in-motion, before it reaches "outputport" stage).
Since DQX natively supports ODCS and there is already a [method to generate DQX rules from an ODCS compliant datacontract](https://databrickslabs.github.io/dqx/docs/reference/api/datacontract/contract_rules_generator/#generate_rules_from_contract) , I am not sure if there is a huge benefit of having this conversion also in the datacontract-cli. For executing the checks you anyways need to have the dqx libraries onboarded, hence we could use the existing method already.......or am I missing something here?

simonharrer:
@gkoenig perhaps have a look at the dbt integration, with the new dbt sync command (which will be improved along the way as well). But the idea could be to simply have "$ datacontract dqx sync", which would make sure that all tests defined within the data cotnract are available as dqx tests, but there can also be additional dqx tests as well (they will be merged together), run all the dqx tests, and convert the test results in the test result format of the data contract cli, and possibly report them back to a system that understands that format.

gkoenig:
> @gkoenig perhaps have a look at the dbt integration, with the new dbt sync command (which will be improved along the way as well). But the idea could be to simply have "$ datacontract dqx sync", which would make sure that all tests defined within the data cotnract are available as dqx tests, but there can also be additional dqx tests as well (they will be merged together), run all the dqx tests, and convert the test results in the test result format of the data contract cli, and possibly report them back to a system that understands that format.

thanks @simonharrer for getting back on this.
I got the point, that such a "sync" method would be a simple solution while keeping the datacontract-cli as lean as possible.
On the other hand, I still don't see the benefit of having a "dqx sync", which would make tests from the ODCS contract available for dqx....since dqx can natively read the ODCS contract and pick the tests from there directly.
What would be the added value of having an additional step in between ?

Since this discussion is now somehow no longer related to the MR changes, I will close this MR, ok ?

jochenchrist:
Agree to close this MR for now. 
Also with respect to v1.0.0, where we switched from Soda v3 to a native engine using ibis as an abstraction layer.

Still open to discuss, how we could better integrate dqt.

PR #1085, opened 2026-03-19
feat: add --schema-name option to test command

### LINKED ISSUE #1079: [Feature Request] Add --schema-name to test to filter
Hello,

it would be helpful to add a `--schema-name`  to `test` similar to that of `export` to select which object we would like to test instead of testing everything.

kelsoufi-sanofi:
Closes #1079

## Summary
- Adds a `--schema-name` option to the `test` command, allowing users to test a specific schema (e.g., `--schema-name orders`) instead of all schemas
- Defaults to `"all"`, preserving existing behavior
- Includes unit tests for schema filtering and a CLI help test

## Test plan
- [x] All existing tests pass (`pytest`)
- [x] New tests cover: filtering by name, default behavior, non-existent schema
- [x] CLI help shows `--schema-name` option
- [x] `ruff format` and `ruff check` pass clean

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## Unreleased
 
+### Added
+- Added `--schema-name` option to `test` command to test a specific schema instead of all schemas (#1079 @kelsoufi-sanofi)
+
 ### Fixed
 - Fix Protobuf export for arrays of objects and improve message/enum naming to UpperCamelCase (#1012 @Schokuroff)
 


datacontract/cli.py
```diff
@@ -177,7 +181,7 @@ def test(
         publish_url=publish,
         server=server,
         ssl_verification=ssl_verification,
-    ).test()
```
jochenchrist, line 202: please do not add the parameter to the method signature. Please keep it in line with the other config options.

kelsoufi-sanofi: done

datacontract/engines/data_contract_test.py
jschoedl: But this will still run for all schemas, or did I overlook something?

tests/test_cli.py
```diff
@@ -16,3 +16,10 @@ def test_file_does_not_exist():
     result = runner.invoke(app, ["test", "unknown.yaml"])
     assert result.exit_code == 1
     assert "The file 'unknown.yaml' does not \nexist." in result.stdout
+
+
+def test_test_schema_name_option_in_help():
+    """Test that --schema-name option is available in test command help."""
+    result = runner.invoke(app, ["test", "--help"])
```
jschoedl:
```suggestion
    result = runner.invoke(app, ["test", "--help"], env={"COLUMNS": "200"})
```

Seems like the test output gets truncated in the CI test, making `--schema-name` not fully appear there. Increasing the terminal width for this specific test could help.

datacontract/engines/data_contract_test.py
kelsoufi-sanofi: by default yes, to stay backward compatible.

jschoedl: Yeah, but if one is specified? The schema_name filter is applied in create_checks, but check_jsonschema (line 58) and check_soda_execute (line 59) in data_contract_test.py still iterate over all schemas.

kelsoufi-sanofi: @jschoedl good catch.
ran a deep search to see where else this needs to be propagated

kelsoufi-sanofi:
hello @jochenchrist @jschoedl 
thanks for the feedback.
let me know if im still missing some coverage for the feature.

PR #1092, opened 2026-03-22
fix: generate composite PRIMARY KEY constraint for multi-column PKs in SQL export

### LINKED ISSUE #1026: Multiple PRIMARY KEY Constraints Generated for Composite Keys in SQL Export
When exporting a data contract with a composite primary key to SQL DDL, datacontract-cli incorrectly generates a separate `PRIMARY KEY` constraint for each column in the composite key, rather than generating a single composite `PRIMARY KEY` constraint.

## Context

This issue arose during the migration from the legacy datacontract format to ODCS (Open Data Contract Standard). In the legacy format, there were two ways to define primary keys:
- On the field level (for single-column keys)
- On the model level (for composite keys)

ODCS changed this to a unified approach where:
- `primaryKey: true` is always specified at the property level
- `primaryKeyPosition` defines the order within a composite key

However, the SQL export code was not updated to handle this properly and continues treating each `primaryKey: true` property as an independent primary key constraint.

## Current Behavior

Given this ODCS contract:

```yaml
version: 0.1.1
kind: DataContract
apiVersion: v3.1.0
id: urn:datacontract:example:composite-key-test
name: Composite Key Example
status: active
schema:
- name: MY_TABLE
  physicalType: table
  properties:
  - name: id_part1
    physicalType: varchar(50)
    logicalType: string
    required: true
    primaryKey: true
    primaryKeyPosition: 1
  - name: id_part2
    physicalType: varchar(50)
    logicalType: string
    required: true
    primaryKey: true
    primaryKeyPosition: 2
  - name: data_field
    physicalType: varchar(256)
    logicalType: string
```

The generated SQL DDL is:

```sql
CREATE TABLE MY_TABLE (
  id_part1 VARCHAR(50) PRIMARY KEY,
  id_part2 VARCHAR(50) PRIMARY KEY,
  data_field VARCHAR(256)
);
```

This is invalid SQL in most databases (you cannot have multiple PRIMARY KEY constraints on a single table).

## Expected Behavior

The generated SQL should be:

```sql
CREATE TABLE MY_TABLE (
  id_part1 VARCHAR(50) NOT NULL,
  id_part2 VARCHAR(50) NOT NULL,
  data_field VARCHAR(256),
  CONSTRAINT pk_MY_TABLE PRIMARY KEY (id_part1, id_part2)
);
```

Where:
- Individual columns are marked as `NOT NULL` (since they're part of the primary key)
- A single table-level `PRIMARY KEY` constraint is generated listing all columns in order of their `primaryKeyPosition`
- The constraint is named following the convention `pk_{table_name}` for better maintainability

For a single-column primary key, the output could be either:

```sql
CREATE TABLE MY_TABLE (
  id VARCHAR(50) NOT NULL,
  data_field VARCHAR(256),
  CONSTRAINT pk_MY_TABLE PRIMARY KEY (id)
);
```

Or continue using the column-level modifier for backward compatibility:

```sql
CREATE TABLE MY_TABLE (
  id VARCHAR(50) PRIMARY KEY,
  data_field VARCHAR(256)
);
```

## Steps to Reproduce

1. Create a data contract with multiple properties having `primaryKey: true` and `primaryKeyPosition` values
2. Run `datacontract export --format sql`
3. Observe multiple `PRIMARY KEY` modifiers in the output

## Additional Notes

The fix should:
- Detect when multiple properties have `primaryKey: true`
- Sort them by `primaryKeyPosition`
- Generate a table-level composite PRIMARY KEY constraint instead of column-level constraints
- Use named constraints (e.g., `CONSTRAINT pk_{table_name}`) for better production use

## Environment

- datacontract-cli version: starting from 0.11.1 up to the current version (0.11.3)
- Output format: SQL DDL

barry0451:
## Problem

When multiple properties have `primaryKey: true`, the SQL exporter emits a separate inline `PRIMARY KEY` per column, which is invalid SQL:

```sql
CREATE TABLE MY_TABLE (
  id_part1 VARCHAR(50) PRIMARY KEY,
  id_part2 VARCHAR(50) PRIMARY KEY,
  data_field VARCHAR(256)
);
```

## Fix

This is a single-topic fix for the composite PK case (related: #1032 addresses this alongside other changes).

- **Single PK:** existing inline `primary key` behavior unchanged
- **Composite PK:** inline `primary key` is suppressed; a table-level `CONSTRAINT pk_{table} PRIMARY KEY (col1, col2)` is appended instead, ordered by `primaryKeyPosition` (nulls sort last)

```sql
CREATE TABLE MY_TABLE (
  id_part1 VARCHAR(50) not null,
  id_part2 VARCHAR(50) not null,
  data_field VARCHAR(256),
  CONSTRAINT pk_MY_TABLE PRIMARY KEY (id_part1, id_part2)
);
```

## Tests

- `test_to_sql_ddl_composite_primary_key`: asserts correct table-level constraint and no inline `primary key`
- `test_to_sql_ddl_single_primary_key`: regression guard for single-PK behavior

Fixes #1026

Co-authored-by: Garrett Sutula <11978161+garrettsutula@users.noreply.github.com>

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -329,7 +329,9 @@ fabric.properties
 
 # Generated test files
 **/.tmp
-/.claude/settings.local.json
+
+# Claude Code
+.claude/
 
 # Auto Claude data directory
 .auto-claude/
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -89,12 +89,6 @@ datacontract export --format html datacontract.yaml --output datacontract.html
 
 # Import from a different format
 datacontract import --format sql --source my-ddl.sql --dialect postgres --output datacontract.yaml
-
-# Find differences between data contracts
-datacontract diff datacontract-v1.yaml datacontract-v2.yaml
-
-# Check for breaking changes
-datacontract breaking datacontract-v1.yaml datacontract-v2.yaml
 ```
 
 ## Project Architecture
@@ -116,8 +110,6 @@ The Data Contract CLI is an open-source command-line tool for working with data 
    - `datacontract/imports/`: Importers from formats like SQL, Avro, JSON Schema, etc.
 
 5. **Linting (`datacontract/lint/`)**: Tools for validating data contract files against schema and best practices.
-
-6. **Breaking Change Detection (`datacontract/breaking/`)**: Logic for identifying breaking changes between versions.
 
 ### Extension Pattern
 
--- a/datacontract/cli.py
+++ b/datacontract/cli.py
@@ -62,7 +62,7 @@ def common(
 
     It uses data contract YAML files to lint the data contract,
     connect to data sources and execute schema and quality tests,
-    detect breaking changes, and export to different formats.
+    and export to different formats.
     """
     pass
 
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -154,7 +154,7 @@ dev = [
   "pytest",
   "pytest-xdist",
   "pymssql==2.3.13",
-  "ruff==0.15.2",
+  "ruff==0.15.7",
   "testcontainers[minio,postgres,kafka,mssql,mysql]==4.14.1",
   "trino==0.336.0",
 ]


datacontract/export/sql_exporter.py
```diff
@@ -118,16 +122,25 @@ def _to_sql_table(model_name: str, model: SchemaObject, server_type: str = "snow
...
             result += " not null"
-        if prop.primaryKey:
+        if prop.primaryKey and not composite_pk:
             result += " primary key"
         if server_type == "databricks" and prop.description is not None:
             result += f' COMMENT "{_escape(prop.description)}"'
         if server_type == "snowflake" and prop.description is not None:
             result += f" COMMENT '{_escape(prop.description)}'"
-        if current_field_index < fields:
+        if current_field_index < total_lines:
```
jschoedl: current_field_index could be renamed to current_line.

tests/test_export_sql.py
```diff
@@ -84,6 +84,32 @@ def test_to_sql_ddl_physical_name():
...
+);
+""".strip()
+    assert actual == expected
+    # Ensure no inline "primary key" per column
+    assert " primary key" not in actual.lower().replace("constraint pk_", "").replace("primary key (", "pk_(")
+
+
+def test_to_sql_ddl_single_primary_key():
+    """Single PK should still use inline primary key (existing behavior)."""
+    actual = DataContract(data_contract_file="fixtures/postgres-export/datacontract.yaml").export("sql")
```
jschoedl: This data contract doesn't have any primary key set. This case is still nice to test, but please add a test with a fixture having exactly one PK, and verify that it actually sets the `primary key` modifier.

tests/test_export_sql.py
```diff
@@ -84,6 +84,32 @@ def test_to_sql_ddl_physical_name():
...
+  order_id integer not null,
+  product_id integer not null,
+  quantity integer,
+  created_at timestamptz,
+  CONSTRAINT pk_orders PRIMARY KEY (order_id, product_id)
+);
+""".strip()
+    assert actual == expected
+    # Ensure no inline "primary key" per column
+    assert " primary key" not in actual.lower().replace("constraint pk_", "").replace("primary key (", "pk_(")
```
jschoedl: Why is this required? If `actual == expected`, everything should be fine already.

tests/fixtures/composite-pk-export/datacontract.yaml
```diff
@@ -0,0 +1,30 @@
...
+    database: test
+    schema: public
+models:
+  orders:
+    type: table
+    fields:
+      order_id:
+        type: integer
+        primaryKey: true
+        primaryKeyPosition: 1
```
jschoedl: If you swap 1 and 2, we would also test for the correct sorting behavior (instead of just using the order in the file). Also, please add a field without `primaryKeyPosition`.

datacontract/export/sql_exporter.py
```diff
@@ -112,25 +112,38 @@ def _to_sql_table(model_name: str, model: SchemaObject, server_type: str = "snow
...
-    current_field_index = 1
+    # When using a table-level composite PK constraint, we need an extra line for it
+    total_lines = fields + (1 if composite_pk else 0)
+    current_line = 1
 
     for prop in properties:
         type_str = convert_to_sql_type(prop, server_type)
         column_name = prop.physicalName or prop.name
         result += f"  {column_name} {type_str}"
         if prop.required:
```
mfurak, line 126:
```suggestion
        if prop.required or prop.primaryKey:
```
As both [PostgreSQL Constraints](https://www.postgresql.org/docs/current/ddl-constraints.html#DDL-CONSTRAINTS-PRIMARY-KEYS) and [Databricks Constraints](https://docs.databricks.com/aws/en/sql/language-manual/sql-ref-syntax-ddl-create-table-constraint#parameters) implicitly create the `NOT NULL` condition, would it make sense to make it explicit when generating code? Additionally, it could make sense to also extend the validation to include a check that when a field has `primaryKey: true` it must also include `required: true` in the configuration.

jschoedl: Good point! This is not directly related to this PR, right? I'd suggest you open a new issue and/or PR on this.

tests/test_export_sql.py
```diff
@@ -84,6 +84,45 @@ def test_to_sql_ddl_physical_name():
     assert actual == expected
 
 
+def test_to_sql_ddl_composite_primary_key():
+    """Composite PKs should generate a table-level CONSTRAINT, not inline per-column primary key.
+
+    Fixture has: order_id at primaryKeyPosition 2, product_id at position 1, note with no position.
+    Note: primaryKeyPosition is not preserved through the ODCS schema_ pipeline, so fields appear
+    in definition order. The fixture still exercises the path via the positionless 'note' field.
```
jschoedl: I see. `primaryKeyPosition` is actually not supported for the (old) Data Contract Standard. I used the CLI to convert the data contract to the new ODCS format which supports it.

PR #1093, opened 2026-03-22
fix: preserve parameterized physicalType in SQL type converter

### LINKED ISSUE #1086: SQL Type Converter cannot handle precision/length contraints for physicalType
The sql_type_converter.py currently maps all string-related types (string, varchar, text) to a generic text type. This ignores specific length requirements defined in physicalType (e.g., VARCHAR(255) converts to None), although the ODCS standard supports these types for physicalTypes of properties.

**Suspected Code Location:**
In sql_type_converter.py:
```
# Current implementation:
if type.lower() in ["string", "varchar", "text"]:
    if format == "uuid":
        return "uuid"
    return "text" # Always returns "text", ignoring the original physicalType or length
```

**Expected Behavior:**

- If a physicalType like VARCHAR(255) is provided, it should be respected.
- If logicalType is string and a maxLength is provided, it should generate VARCHAR(length).
- Only if no specific length is provided should it fall back to text or VARCHAR.

barry0451:
## Problem

When an ODCS `SchemaProperty` has a `physicalType` with precision or length constraints (e.g. `VARCHAR(255)`, `DECIMAL(10,2)`), the SQL type converter ignores the constraints and returns the wrong type or `None`.

**Root cause:** The per-server converters match bare type names like `["string", "varchar", "text"]`. A `physicalType` of `"VARCHAR(255)"` doesn't match `"varchar"`, so the converter falls through and returns `None`.

## Fix

Added an early return in `convert_to_sql_type()`: if the ODCS `physicalType` already contains parentheses, it is returned verbatim before any server-specific converter runs.

Also adds `_extract_base_type()` helper (strips params from a type string) for future use.

## Tests

63 parametrized tests covering:
- `VARCHAR(255)`, `DECIMAL(10,2)`, `NUMERIC(18,4)`, `CHAR(10)`, `NVARCHAR(500)`, `FLOAT(24)` across all 8 server types
- Regression tests confirming bare types still route through server-specific converters correctly

Fixes #1086

Co-authored-by: Garrett Sutula <11978161+garrettsutula@users.noreply.github.com>

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 - Fix SQL export generating multiple PRIMARY KEY constraints for composite keys (#1026)
+- Preserve parametrized physicalTypes for SQL export (#1086)
 
 ## [0.11.7] - 2026-03-24
 
--- a/datacontract/export/sql_type_converter.py
+++ b/datacontract/export/sql_type_converter.py
@@ -123,7 +123,10 @@ _TYPES_ACCEPTING_PARAMS: Dict[str, FrozenSet[str]] = {
             "number",
             "decimal",
             "numeric",
-            "float",
+            "binary",
+            "timestamp_tz",
+            "timestamp_ntz",
+            "time",
         ]
     ),
     "postgres": frozenset(
@@ -136,6 +139,9 @@ _TYPES_ACCEPTING_PARAMS: Dict[str, FrozenSet[str]] = {
             "decimal",
             "bit",
             "bit varying",
+            "timestamptz",
+            "timestamp",
+            "time",
         ]
     ),
     "databricks": frozenset(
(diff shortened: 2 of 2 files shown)


barry0451:
Thanks for the feedback — good catch.

The naive verbatim pass-through is replaced with proper server-aware translation:

1. Strip parameters from the physicalType (e.g. `NVARCHAR(50)` → base=`NVARCHAR`, params=`50`)
2. Run the base type through the server-specific converter
3. Re-attach params only if the translated type accepts them on that server (via a `_TYPES_ACCEPTING_PARAMS` per-server mapping)

Examples of the new behaviour:
- `NVARCHAR(50)` on Postgres → converter returns `None` for unknown base type → fallback verbatim `NVARCHAR(50)` (Postgres will reject this cleanly rather than silently producing wrong SQL)
- `VARCHAR(255)` on Postgres → base `varchar` → `text`, text doesn't accept params → `text`
- `DECIMAL(10,2)` on Postgres → base `decimal` → `decimal`, accepts params → `decimal(10,2)`
- `VARCHAR(255)` on Snowflake → base `varchar` → `VARCHAR`, accepts params → `VARCHAR(255)`

21 new tests cover helpers and cross-server translation cases.

PR #1112, opened 2026-03-31
feat(diff): add ODCS semantic diff command and API endpoint

davidb-tada:
# feat(diff): semantic diff for ODCS data contracts with text and HTML reports

## Summary
Reintroducing the much loved diff functionality, with a wider capability including semantic diff with field level granularity. This PR adds a `datacontract diff` command that produces a semantic diff of two data contract YAML files, with plain-text and HTML report formats.

**Note:** Breaking change classification (which changes are breaking vs non-breaking) is out of scope for this PR and will follow in a subsequent one. 

```bash
# Plain-text diff to stdout
datacontract diff v1.odcs.yaml v2.odcs.yaml

# HTML report saved to a file
datacontract diff v1.odcs.yaml v2.odcs.yaml --format html --output diff.html
```

A `POST /diff` endpoint is also added to the API server, accepting the two contracts as YAML strings in the request body.

## Motivation
When evolving a data contract, understanding what changed between two versions is a prerequisite to assessing impact. This provides a structured, human-readable diff rather than a raw YAML text diff, with changes grouped by field path and presented at both a summary (rolled-up) and detail (leaf) level.

### HTML Report
golden_integration_html

### Text Report
golden_text_report

## How it works
The diff engine normalises both contracts before diffing — converting named lists (e.g. `schema[]`, `customProperties[]`, `authoritativeDefinitions[]`) from positional arrays to dicts keyed by their natural identifier. 

This eliminates false positives from list reordering and produces stable, meaningful field paths in the output e.g. 
`schema.orders.properties.order_id.logicalType        Changed` 
rather than 
`schema[0].properties[1].logicalType                  Changed`

The renderer uses `LIST_CONTAINERS` to apply list-item prefixes (`- `) to named collection entries in the output. This set is derived automatically by reflecting over the ODCS Pydantic models, so it stays in sync with the schema without manual maintenance.

## Natural key limitation and upstream path - Future Considerations
 
This section provides context around the current semantic / natural key implementation and envisaged improvements.

The natural key for each named list — the field used to key items when converting arrays to dicts before diffing — is currently hardcoded in `report_helpers.py`. Each entry in the table below was determined by inspecting the `required` array in the [ODCS v3 JSON Schema](https://github.com/bitol-io/open-data-contract-standard/blob/main/schema/odcs-json-schema-v3.1.0.json) and selecting the field that acts as the stable semantic identifier:
 
| Entity | Natural key | Source |
|---|---|---|
| `SchemaObject` | `name` | `required: [name]` in JSON schema |
| `SchemaProperty` | `name` | `required: [name]` in JSON schema |
| `Server` | `server` | `required: [server, type]` — `server` chosen as identifier |
| `SLAProperty` | `property` | `required: [property, value]` — `property` chosen as identifier |
| `CustomProperty` | `property` | `required: [property, value]` — `property` chosen as identifier |
| `Role` | `role` | `required: [role]` in JSON schema |
| `SupportItem` | `channel` | `required: [channel]` in JSON schema |
| `TeamMember` | `username` | `required: [username]` in JSON schema |
| `DataQuality` | `name` | no `required` in JSON schema — `name` inferred; positional fallback if absent |
 
The `DataQuality` case illustrates the root problem: the JSON schema does not declare any field as `required` for that object, so there is no machine-readable source of truth from which the natural key can be derived. The key must instead be inferred from domain knowledge and hardcoded with a fallback.
 
The correct long-term fix is for the natural key to be **declared in the spec itself** — specifically, by ensuring that each named-list entity type in [`odcs-json-schema-v3.1.0.json`](https://github.com/bitol-io/open-data-contract-standard/blob/main/schema/odcs-json-schema-v3.1.0.json) has a `required` array whose first entry is unambiguously its stable identifier. That metadata would then flow through to the [Pydantic model package](https://github.com/datacontract/open-data-contract-standard-python) as a non-optional field, allowing `helpers.py` to derive the natural key table dynamically by reflecting over `model_fields` rather than maintaining it by hand.
 
If this PR is merged, the author intends to raise issues and submit PRs to both upstream projects — [`bitol-io/open-data-contract-standard`](https://github.com/bitol-io/open-data-contract-standard) to add the missing `required` declarations to the JSON schema, and [`datacontract/open-data-contract-standard-python`](https://github.com/datacontract/open-data-contract-standard-python) to surface those as non-optional Pydantic fields — which would allow a follow-up PR here to simplify the implementation, replace the hardcoded key table and its associated logic with fully derived logic. 

## Changes
- `datacontract/reports/diff/diff.py` — ODCS semantic diff implementation with diff pipeline
- `datacontract/reports/diff/contract_diff_report.py` — builds the report data dict from raw diff output
- `datacontract/reports/diff/text_contract_diff_renderer.py` — self-contained plain-text renderer
- `datacontract/reports/diff/html_contract_diff_renderer.py` — self-contained HTML renderer
- `datacontract/reports/common/report_helpers.py` — helper code to facilitate semantic diffs
- `datacontract/reports/common/html_base_renderer.py` — common HTML base rendering functionality
- `datacontract/api.py` — `POST /diff` endpoint
- `datacontract/cli.py` — `datacontract diff` command
- `tests/test_reports_diff_*.py` — comprehensive test suite covering diff, report building, rendering, and API
- `tests/test_reports_common_helpers.py` — tests for common report helper functions
- `tests/fixtures/diff/*` — new fixture files for ODCS diff testing
- `tests/fixtures/breaking/*` — removed leftover fixture files from the DCS based implementation
- `API.md` — diff endpoint documentation
- `CHANGELOG.md` — unreleased entry
- `README.md` — updated with new diff functionality
- `pyproject.toml` — dependency updates

## Testing
```bash
uv run pytest tests/test_rep* tests/test_cli.py tests/test_api.py
```
266 tests, all passing.

- [x] Tests pass
- [x] ruff format

### DIFF
--- a/datacontract/reports/common/html_base_renderer.py
+++ b/datacontract/reports/common/html_base_renderer.py
@@ -7,16 +7,12 @@ ODCS report renderers. It includes CSS styling, HTML escaping, formatting
 utilities, and base renderer classes.
 
 NOTE: The functions in this module are tested indirectly through the HTML renderer tests.
-The module-level functions (_e, _pill, _path_td, _format_value, etc.) are re-exported by
-HtmlContractDiffRenderer and tested in tests/test_reports_diff_html_renderer.py.
-This approach encapsulates the base renderer as an internal implementation detail while
-ensuring thorough testing of the functionality in the context where it's actually used.
 
 """
 
 from __future__ import annotations
 
-import html as _html
+import html
 from typing import Any
 
 from datacontract.reports.common.report_helpers import LIST_CONTAINERS, _infer_ancestors  # noqa: F401 — re-exported
@@ -157,12 +153,12 @@ _CSS = _css()
 # fragments (a string, a span, a <td>) and are imported directly by renderers.
 # ---------------------------------------------------------------------------
 
-def _e(s: Any) -> str:
+def e(s: Any) -> str:
     """HTML-escape a value."""
-    return _html.escape(str(s))
+    return html.escape(str(s))
 
 
-def _format_value(val: Any) -> str:
+def format_value(val: Any) -> str:
     """Format a cell value for display in the diff table.
 
     dict   → key summary span:  { key1, key2, ... }
@@ -174,11 +170,11 @@ def _format_value(val: Any) -> str:
     if isinstance(val, dict):
         keys = list(val.keys())
         summary = f"{{ {', '.join(keys[:4])}{'...' if len(keys) > 4 else ''} }}"
-        return f'<span class="obj">{_e(summary)}</span>'
-    return _e(str(val))
+        return f'<span class="obj">{e(summary)}</span>'
+    return e(str(val))
 
 
-def _pill(change_type: str) -> str:
+def pill(change_type: str) -> str:
     """Render a coloured change-type pill span.
 
     Accepts 'added', 'removed', or 'changed'. Unknown types are capitalised
@@ -191,7 +187,7 @@ def _pill(change_type: str) -> str:
     return f'<span class="pill {css}">{label}</span>'
 
 
-def _path_td(key: str, depth: int, is_list_item: bool) -> str:
+def path_td(key: str, depth: int, is_list_item: bool) -> str:
     """Build a depth-indented <td> for the path column.
 
     depth        — number of ancestor segments; drives the padding-left calc
@@ -199,7 +195,7 @@ def _path_td(key: str, depth: int, is_list_item: bool) -> str:
                    causing a "- " prefix so the item reads as a list entry
     """
     pad = f"padding-left:calc(14px + {depth * 2}ch)"
-    label = f"- {_e(key)}" if is_list_item else _e(key)
+    label = f"- {e(key)}" if is_list_item else e(key)
     return f'<td class="path" style="{pad}"><span class="key">{label}</span></td>'
 
 
@@ -248,9 +244,9 @@ class HtmlContractBaseRenderer:
         return f"""
         <div class="header">
           <div class="header-left">
-            <h1>{_e(title)}</h1>
-            <div class="subtitle">{_e(subtitle)}</div>
-            <div class="subtitle" style="margin-top:4px;">Generated: {_e(generated_at)}</div>
+            <h1>{e(title)}</h1>
+            <div class="subtitle">{e(subtitle)}</div>
+            <div class="subtitle" style="margin-top:4px;">Generated: {e(generated_at)}</div>
           </div>
         </div>"""
 
@@ -274,7 +270,7 @@ class HtmlContractBaseRenderer:
 <head>
 <meta charset="UTF-8">
 <meta name="viewport" content="width=device-width, initial-scale=1.0">
-<title>{_e(page_title)} — {_e(source_label)} → {_e(target_label)}</title>
+<title>{e(page_title)} — {e(source_label)} → {e(target_label)}</title>
 <style>{_CSS}</style>
 </head>
 <body>
--- a/datacontract/reports/diff/html_contract_diff_renderer.py
+++ b/datacontract/reports/diff/html_contract_diff_renderer.py
@@ -12,11 +12,11 @@ from __future__ import annotations
 from datacontract.reports.common.html_base_renderer import (
     _CSS,  # noqa: F401 — re-exported for any callers that import it directly
     HtmlContractBaseRenderer,
-    _e,
-    _format_value,
     _infer_ancestors,
-    _path_td,
-    _pill,
+    e,
+    format_value,
+    path_td,
+    pill,
 )
 from datacontract.reports.common.report_helpers import LIST_CONTAINERS
 
@@ -25,10 +25,10 @@ from datacontract.reports.common.report_helpers import LIST_CONTAINERS
 # import surface of html_report_renderer is unchanged.
 __all__ = [
     "HtmlContractDiffRenderer",
-    "_e",
-    "_format_value",
-    "_path_td",
-    "_pill",
+    "e",
+    "format_value", 
+    "path_td",
+    "pill",
     "_render_detail_rows",
 ]
 
@@ -53,24 +53,24 @@ def _render_detail_rows(changes: list) -> list[str]:
         key = segs[-1]
         depth = len(segs) - 1
         is_list_item = len(segs) > 1 and segs[-2] in LIST_CONTAINERS
-        td = _path_td(key, depth, is_list_item)
+        td = path_td(key, depth, is_list_item)
 
         if c.get("_ancestor"):
             rows.append(f"<tr>{td}<td></td><td></td><td></td></tr>")
             continue
 
         change_type = c["changeType"].lower()
-        pill = _pill(change_type)
+        pill_html = pill(change_type)
         old_v = c.get("old_value")
         new_v = c.get("new_value")
 
         if old_v is None and new_v is None:
-            rows.append(f"<tr>{td}<td>{pill}</td><td></td><td></td></tr>")
+            rows.append(f"<tr>{td}<td>{pill_html}</td><td></td><td></td></tr>")
         else:
-            old_html = _format_value(old_v) if old_v is not None else ""
-            new_html = _format_value(new_v) if new_v is not None else ""
+            old_html = format_value(old_v) if old_v is not None else ""
+            new_html = format_value(new_v) if new_v is not None else ""
             rows.append(
-                f'<tr>{td}<td>{pill}</td><td class="new">{new_html}</td><td class="old">{old_html}</td></tr>'
+                f'<tr>{td}<td>{pill_html}</td><td class="new">{new_html}</td><td class="old">{old_html}</td></tr>'
             )
 
     return rows
@@ -119,8 +119,8 @@ class HtmlContractDiffRenderer(HtmlContractBaseRenderer):
 
         rows = []
         for ch in changes:
-            pill = _pill(ch["changeType"].lower())
-            rows.append(f'<tr><td class="path">{_e(ch["path"])}</td><td>{pill}</td></tr>')
+            pill_html = pill(ch["changeType"].lower())
+            rows.append(f'<tr><td class="path">{e(ch["path"])}</td><td>{pill_html}</td></tr>')
 
         total = counts["added"] + counts["removed"] + counts["changed"]
         summary_header = self._render_summary_html(total, badges_html)
--- a/tests/test_reports_diff_html_renderer.py
+++ b/tests/test_reports_diff_html_renderer.py
@@ -2,100 +2,98 @@
 test_html_report_renderer — Unit tests for HtmlContractDiffRenderer
 ------------------------------------------------------------------
 Test classes:
-    TestHtmlEscape       — _e(): HTML escaping
-    TestFormatValue      — _format_value(): dict summary, truncation, escaping
-    TestPill             — _pill(): change-type pill markup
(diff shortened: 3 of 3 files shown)


datacontract/reports/common/__init__.py
jochenchrist, line 1: I am not a big fan of "common" packages. Mostly there is not much in common. Please move the logic to the actual reports.

tests/fixtures/diff/integration/golden_integration_html.html
```diff
@@ -0,0 +1,149 @@
+<!DOCTYPE html>
+<html lang="en">
+<head>
+<meta charset="UTF-8">
+<meta name="viewport" content="width=device-width, initial-scale=1.0">
+<title>ODCS Diff — tests/fixtures/diff/integration/report_renderer_integration_v1.yaml → tests/fixtures/diff/integration/report_renderer_integration_v2.yaml</title>
+<style>
+@import url('https://fonts.googleapis.com/css2?family=IBM+Plex+Mono:wght@400;500&family=IBM+Plex+Sans:wght@300;400;500;600&display=swap');
```
jochenchrist, line 8: Why need for this font? Is it used anywhere else?

datacontract/reports/common/__init__.py
davidb-tada: OK - The intent was not explicit here: I am laying the foundation for the next PR that will focus on breaking changes.

Both `diff` and `breaking` will be sharing that rendering logic, report layout and UX, ensuring they are aligned on these aspects. This is the reason `common` is being introduced here.

tests/fixtures/diff/integration/golden_integration_html.html
davidb-tada: This font is used as part of the html diff report, and will also be used by the breaking report.

This font provides a good look-and-feel (cf screenshot in the PR  description).

datacontract/reports/common/__init__.py
jschoedl: For now, I think they fit better in the `reports` directory directly. This makes it similarly obvious that they are not specific to either `diff` or `breaking`. 2 shared files are not really a reason for a common package.

Also, there are many functions that are only called in one of `html_contract_diff_renderer` or `text_contract_diff_renderer` (and in the tests). Please move those to the place where they are called, as @jochenchrist said. Only when they are actually used for the `breaking` logic, they should be moved into a third file.

datacontract/api.py
```diff
@@ -358,6 +367,56 @@ async def lint(
...
+    try:
+        contracts_diff = ContractDiff().generate(v1_path, v2_path)
+        report_data = ContractDiffReport().build_report_data(contracts_diff, source_label="v1", target_label="v2")
+        if format == "html":
+            content = HtmlContractDiffRenderer(report_data=report_data).render()
+            return HTMLResponse(content=content)
+        else:
+            content = TextContractDiffRenderer(report_data=report_data).render()
+            return PlainTextResponse(content=content)
+    finally:
```
jschoedl, line 421:
```suggestion
    except yaml.YAMLError as e:
        raise HTTPException(status_code=422, detail=f"Invalid YAML: {e}")
    except ValidationError as e:
        raise HTTPException(status_code=422, detail=f"Invalid data contract: {e}")
    finally:
```

Maybe make this a bit more fine-grained, as an invalid contract should not throw http code 500. (I noticed that we're currently returning 500 for other commands as well, might be worth a future change.)

datacontract/reports/diff/contract_diff_report.py
```diff
@@ -0,0 +1,231 @@
...
+                    entries.append(entry)
+            return entries
+
+        detail_changes = []
+        for deepdiff_key, items in diff_result.items():
+            change_type = self.CHANGE_TYPE_MAP.get(deepdiff_key)
+            if not change_type:
+                continue
+            for raw_path, payload in items.items():
+                segs = re.findall(r"\['([^']+)'\]", raw_path)
```
jschoedl:
```suggestion
                # match ['key'] or ["key"]
                segs = re.findall(r"""(?:\['([^']+)'\]|\["([^"]+)"\])""", raw_path)
```
Although unlikely, it might happen that keys contain `'`, in which case they will be quoted using `"`. Let's make sure double quotes work as well.

datacontract/reports/diff/diff.py
```diff
@@ -0,0 +1,318 @@
...
+            _positional.append(_a)
+    if len(_positional) != 2:
+        print("Usage: python diff.py <v1.yaml> <v2.yaml> [--html] [--output <file>]")
+        sys.exit(1)
+    ContractDiffReport().generate(
+        v1_path=_positional[0],
+        v2_path=_positional[1],
+        fmt=_fmt,
+        output_path=_output_path,
+    )
```
jschoedl, line 315: Is this a leftover from your testing? We do not have other `__main__` entry points in the codebase except for cli.py.

datacontract/reports/diff/html_contract_diff_renderer.py
```diff
@@ -0,0 +1,163 @@
...
+Produces a self-contained HTML diff report from a report_data diff dict,
+including a Change Summary section and a Change Details table with
+depth-based indentation.
+
+"""
+
+from __future__ import annotations
+
+from datacontract.reports.common.html_base_renderer import (
+    _CSS,  # noqa: F401 — re-exported for any callers that import it directly
```
jschoedl, line 13: this import is not used here, right?

CHANGELOG.md
```diff
@@ -49,7 +26,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Fix mkdir TOCTOU race causing silent JUnit write failure (#1050)
 - Fix validation failure for field names with special chars on Databricks (#1049)
 - Add Azure support for field name quoting in schema checks (#1025)
-
+hint: Resolve all conflicts manually, mark them as resolved with
+hint: "git add/rm <conflicted_files>", then run "git rebase --continue".
+hint: You can instead skip this commit: run "git rebase --skip".
+hint: To abort and get back to the state before "git rebase", run "git rebase --abort".
```
jschoedl, line 53: remove

CHANGELOG.md
```diff
@@ -7,34 +7,11 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
-
-### Fixed
-- Fix parser error for CSV / Parquet table names containing special characters (#1066)
-- Fix BigQuery export failing with "Unsupported type" for parameterized physicalType like `NUMERIC(18, 4)` (#1083)
-
-### Added
 - Added JSON output format for test results (`--output-format json`)
-- Added Azure AD / Entra ID authentication support for SQL Server and Microsoft Fabric
+- Added Azure AD / Entra ID authentication support for SQL Server (`ActiveDirectoryPassword`, `ActiveDirectoryServicePrincipal`, `ActiveDirectoryInteractive`)
+- (feat(diff): Add data contract semantic diff command and API endpoint)
```
jschoedl:
```suggestion
- Add data contract semantic `diff` command and API endpoint
```

CLAUDE.md
```diff
@@ -111,6 +115,10 @@ The Data Contract CLI is an open-source command-line tool for working with data
...
... — planned improvement**: The natural keys used to match list items (e.g. `schema[].name`, `customProperties[].property`) are currently hardcoded in `diff.py` as a 20-line path-to-key table, backed by six helper methods (`_normalize_by`, `_normalize_schema_fields`, `_normalize_quality`, `_normalize_auth_defs`, `_normalize_relationships`, `_normalize_properties`). The intended long-term fix is to encode these natural keys directly in the ODCS JSON Schema — as an `x-natural-key` extension or similar — and carry that through to the upstream `open-data-contract-standard` Pydantic models (via a `__natural_key__` class var or `Field` annotation). This is an improvement beyond just simplifying this code: it makes the identity semantics of each model explicit and authoritative in the schema itself, rather than having them implied by convention or rediscovered by downstream consumers. Once that lands, all six helpers and the key table collapse into a single ~15-line recursive function that walks the model tree by reflection — adding or changing a list field in the ODCS schema would require no changes here. See the `NOTE` in `ContractDiff._normalize()` for the exact upstream change required.
```
jschoedl: Why do we need this here?

tests/fixtures/diff/integration/helper/generate_golden.py
```diff
@@ -0,0 +1,70 @@
...
+
+import os
+from unittest.mock import patch
+
+from datacontract.reports.diff.contract_diff_report import ContractDiffReport
+from datacontract.reports.diff.diff import ContractDiff
+from datacontract.reports.diff.html_contract_diff_renderer import HtmlContractDiffRenderer
+from datacontract.reports.diff.text_contract_diff_renderer import TextContractDiffRenderer
+
+FIXTURE_DIR = os.path.join(os.path.dirname(__file__), "fixtures", "diff", "integration")
```
jschoedl:
```suggestion
FIXTURE_DIR = os.path.join(os.path.dirname(__file__), "..")
```
Invalid path here. I guess you moved the file from project root?

datacontract/reports/common/html_base_renderer.py
```diff
@@ -0,0 +1,285 @@
...
+# as a plain module-level function. These produce small, composable HTML
+# fragments (a string, a span, a <td>) and are imported directly by renderers.
+# ---------------------------------------------------------------------------
+
+def _e(s: Any) -> str:
+    """HTML-escape a value."""
+    return _html.escape(str(s))
+
+
+def _format_value(val: Any, max_len: int = None) -> str:
```
jschoedl: `max_len` is unused. Either remove or implement.

tests/fixtures/diff/integration/golden_integration_html.html
jochenchrist: I would ask to to make the HTML report consistent with the HTML export (following Tailwind Plus styleguide).

We could also remove the HTML report for now to focus the scope of this feature.

davidb-tada: I agree with your suggestion to remove this and reduce the scope for now.

datacontract/reports/diff/html_contract_diff_renderer.py
davidb-tada: Removed - ruff did not detect

datacontract/reports/diff/diff.py
davidb-tada: This was a left over. Removed

CHANGELOG.md
davidb-tada: Removed

datacontract/api.py
davidb-tada: Added, with tests.

CLAUDE.md
davidb-tada: Merged with existing doc in `diff.py`

datacontract/reports/common/html_base_renderer.py
davidb-tada: Removed

davidb-tada:
> > Upon consideration it may be worthwhile aligning at least the text based report output with the existing structure and color scheme.
> > image
> > Any thoughts on this @jochenchrist ? I am happy to look at this if there is commitment on your side to incorporate this when deemed ready.
> 
> I think it is OK when the diff is a bit different to the test results, if contents are different. Of course, from a user perspective, some general UI pattern should be consistent.

Thanks for the review folks.

Based on all feedback I have decided to close this PR and create a new PR that will focus on the text output with an output format and code design that is better aligned with the existing code base.

I will incorporate all the present feedback into the new code

PR #1144, opened 2026-04-13
feat: add --checks option to test command to filter check categories (#678)

### LINKED ISSUE #678: Disable quality checks, only run checks based on schema.
Schema checks are checks are cheap to execute, but are executed at different points in time in comparison to quality checks.

Provide a CLI parameter to disable the quality checks via `--no-qualitychecks`. The are enabled (`--qualitychecks`) by default.

```
datacontract test --no-qualitychecks my-data-contract.yaml
```

jschoedl:
## Summary

- Adds `--checks` option to `datacontract test` to selectively run check categories
- Available categories: `schema`, `quality`, `servicelevel`, `custom`
- Omitting the option runs all categories (default behaviour unchanged)
- Invalid or empty `--checks` values exit with a helpful error message

## Usage

```bash
# Only run schema checks
datacontract test --checks schema datacontract.yaml

# Run schema and quality checks
datacontract test --checks schema,quality datacontract.yaml
```

## Test plan

- [x] `test_checks_schema_only` — only schema-category checks returned
- [x] `test_checks_quality_only_no_quality_checks_defined` — empty result when no quality checks defined
- [x] `test_checks_all_categories_same_as_default` — explicit full list matches default behaviour
- [x] `test_checks_cli_option` — CLI option accepted
- [x] `test_checks_cli_invalid_category` — exits with error on unknown category
- [x] `test_checks_cli_empty_category` — exits with error on empty value
- [x] `test_checks_cli_multiple_categories` — comma-separated list works
- [x] `test_checks_cli_spaces_after_comma` — spaces around commas handled

Closes #678

### DIFF
--- a/datacontract/cli.py
+++ b/datacontract/cli.py
@@ -181,6 +181,14 @@ def test(
         ),
     ] = None,
     output_format: Annotated[OutputFormat, typer.Option(help="The target format for the test results.")] = None,
+    checks: Annotated[
+        str,
+        typer.Option(
+            help="Comma-separated list of check categories to run. "
+            "Available categories: schema, quality, servicelevel, custom. "
+            "Omit to enable all."
+        ),
+    ] = None,
     logs: Annotated[bool, typer.Option(help="Print logs")] = False,
     ssl_verification: Annotated[
         bool,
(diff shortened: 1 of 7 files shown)


datacontract/engines/data_contract_test.py
```diff
@@ -61,12 +62,16 @@ def execute_data_contract_test(
...
+        checks = [c for c in checks if c.category in check_categories]
+    run.checks.extend(checks)
 
     # TODO check server is supported type for nicer error messages
     # TODO check server credentials are complete for nicer error messages
     if server.format == "json" and server.type != "kafka":
-        check_jsonschema(run, data_contract, server, schema_name=schema_name)
+        if check_categories is None or "schema" in check_categories:
+            check_jsonschema(run, data_contract, server, schema_name=schema_name)
     check_soda_execute(run, data_contract, server, spark, duckdb_connection, schema_name=schema_name)
```
kelsoufi-sanofi: we should verify that run.checks is not empty maybe

datacontract/cli.py
```diff
@@ -181,6 +181,14 @@ def test(
         ),
     ] = None,
     output_format: Annotated[OutputFormat, typer.Option(help="The target format for the test results.")] = None,
+    checks: Annotated[
+        str,
+        typer.Option(
+            help="Comma-separated list of check categories to run. "
+            "Available categories: schema, quality, servicelevel, custom. "
```
kelsoufi-sanofi, line 188: is custom captured?

datacontract/engines/data_contract_test.py
jschoedl: I thought that it might be intentional: if you run this command for many contracts, it could be desired that on some of them the CLI does not perform any check. But I'll add a warning in this case.

datacontract/cli.py
jschoedl: Good point, seems like I actually missed filtering custom checks... thank you!

PR #1152, opened 2026-04-14
fix: emit placeholder connection fields in SQL importer server block

### LINKED ISSUE #1146: SQL importer always emits incomplete server stub (missing database/host/port/schema)
## Summary

`datacontract import --format sql` always injects a server block with only `type: postgres`, omitting all required connection fields (`database`, `host`, `port`, `schema`). The output is immediately invalid according to the ODCS schema, yet the import exits 0 with no warning.

## Reproduction

```bash
datacontract import --format sql --source any.sql --dialect postgres --output out.yaml
# EXIT 0 — no warning

datacontract lint out.yaml
# → data.servers[0] must contain ['database', 'host', 'port', 'schema'] properties
# EXIT 1
```

**Generated server section:**
```yaml
servers:
- server: postgres
  type: postgres
  # database, host, port, schema — all required by ODCS, all missing
```

## Expected behaviour

Either:
- ~Prompt for / accept connection details as CLI flags and include them, or~
- **Emit the server block as a clearly commented placeholder so users know they must fill it in, or**
- ~Omit the server block entirely if no connection info is provided (a contract without a server is valid ODCS)~

The current behaviour produces a file that immediately fails `datacontract lint`, giving users a confusing first experience.

Ai-chan-0411:
## Summary

Fixes #1146.

The SQL importer was generating a server stub containing only `type: <dialect>`, omitting all required ODCS connection fields (`host`, `port`, `database`, `schema`). This caused `datacontract lint` to fail immediately after `datacontract import`, giving users a confusing first experience:

```
# Before
servers:
  - server: postgres
    type: postgres
    # database, host, port, schema — all missing, lint fails
```

## Changes

- Added `get_server_defaults(server_type)` that returns sensible placeholder values for each server type, including dialect-appropriate default ports
- Updated `import_sql` to include these placeholders when building the server block, so users can see exactly what fields need to be set
- Added a `logging.warning` to make it clear the values are placeholders
- Updated existing tests to reflect the new expected output

```
# After
servers:
  - server: postgres
    type: postgres
    host: localhost
    port: 5432
    database: database
    schema: public
```

## Test plan

- [x] `test_import_sql_postgres` — updated expected YAML
- [x] `test_import_sql_snowflake` — updated expected YAML
- [x] `test_import_sql_sqlserver` — updated expected YAML
- [x] `test_import_sql_oracle` — updated expected YAML

🤖 Generated with [Claude Code](https://claude.ai/claude-code)

### DIFF
--- a/datacontract/imports/sql_importer.py
+++ b/datacontract/imports/sql_importer.py
@@ -43,7 +43,12 @@ def import_sql(source: str, import_args: dict = None) -> OpenDataContractStandar
 
     server_type = to_server_type(source, dialect)
     if server_type is not None:
-        odcs.servers = [create_server(name=server_type, server_type=server_type)]
+        server_defaults = get_server_defaults(server_type)
+        odcs.servers = [create_server(name=server_type, server_type=server_type, **server_defaults)]
+        logging.warning(
+            "SQL import generated a server block with placeholder connection values. "
+            "Update host, port, database, and schema in the output before use."
+        )
 
     tables = [
         t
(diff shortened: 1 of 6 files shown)


datacontract/imports/sql_importer.py
```diff
@@ -138,6 +143,32 @@ def to_dialect(import_args: dict) -> Dialects | None:
...
+        "mysql": 3306,
+        "sqlserver": 1433,
+        "oracle": 1521,
+        "snowflake": 443,
+        "databricks": 443,
+    }
+    defaults = {
+        "host": "localhost",
+        "database": "database",
+        "schema": "public",
```
jschoedl: Maybe we should use values that make it more obvious that they are invalid values in case the user missed the warning, like "my_host" / "my_database" / "my_schema".

jschoedl (changes requested):
One comment, rest LGTM!

datacontract/imports/sql_importer.py
jochenchrist: This is only true for postgres.

Ai-chan-0411: Good idea! Updated to use my_host / my_database / my_schema as placeholders. Also made schema server-type aware: postgres and redshift keep "public" (their actual default), others get "my_schema".

Ai-chan-0411: Good catch — already addressed in the latest push (bc598c1). I added a `schema_map` so only postgres and redshift get `"public"` as their default schema; all other server types now get `"my_schema"` instead.

PR #1161, opened 2026-04-16
fix: exit with code 1 when --server name is not found

### LINKED ISSUE #1153: test: --server with nonexistent name exits code 0
## Reproduction

```bash
datacontract test --server nonexistent datacontract.yaml
```

Exits with code 0 and a warning. A typo like `--server producton` silently passes CI.

## Expected

Exit code 1 with a clear error:

```
Error: Server 'nonexistent' not found. Available servers: local-test, production
```

Ai-chan-0411:
## Summary

I ran into this while setting up CI for a project with multiple server environments. A typo in the `--server` flag (`producton` instead of `production`) silently passed all checks with exit code 0. The root cause is that `check_that_datacontract_contains_valid_server_configuration` raises a `DataContractException` with `result="warning"` when the server name doesn't match any configured server. Since the CLI only exits with code 1 for non-warning/non-passed results, this means a nonexistent server name produces exit code 0.

### Changes

- Changed the result from `"warning"` to `"failed"` in the server-not-found check, so the CLI exits with code 1
- Improved the error message to list available server names (e.g. `Server 'producton' not found in data contract. Available servers: local-test, production`), matching the pattern already used for the schema-not-found error on line 57 of `data_contract_test.py`
- Added two regression tests: one verifying the `Run` result is `"failed"`, and one verifying the CLI exit code is 1

Closes #1153

## Test plan
- [ ] `test_test_nonexistent_server_exits_with_error` — verifies `run.result == "failed"` and the reason mentions the nonexistent server name
- [ ] `test_test_nonexistent_server_cli_exit_code` — verifies CLI exits with code 1 when given a nonexistent `--server` value

### DIFF
--- a/datacontract/engines/datacontract/check_that_datacontract_contains_valid_servers_configuration.py
+++ b/datacontract/engines/datacontract/check_that_datacontract_contains_valid_servers_configuration.py
@@ -29,8 +29,8 @@ def check_that_datacontract_contains_valid_server_configuration(
             raise DataContractException(
                 type="lint",
                 name="Check that data contract contains valid servers configuration",
-                result="warning",
-                reason=f"Cannot find server '{server_name}' in the data contract servers configuration. Skip executing tests.",
+                result="failed",
+                reason=f"Server '{server_name}' not found in data contract. Available servers: {', '.join(sorted(server_names))}",
                 engine="datacontract",
             )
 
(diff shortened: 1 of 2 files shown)


jschoedl (approved):
LGTM!

PR #1163, opened 2026-04-18
fix: detect missing fields in CSV/Parquet files correctly

### LINKED ISSUE #1065: Check `field_is_present` always passes for CSV/Parquet files

**Possible solutions:**

hieusats:
## Problem

For CSV and Parquet files accessed via DuckDB (local, S3, GCS, Azure), the check `field_is_present` always passes regardless of whether the column exists in the file.

**Root cause:** `create_view_with_schema_union()` creates an empty table with all contract columns, then inserts only intersecting columns from the data. Missing columns remain in the table (filled with NULLs), so SodaCL's `when required column missing` check sees the column as present.

## Fix

1. **Create a `{model}_raw` view** alongside the unioned table that exposes only actual data columns (no contract-only columns)
2. **Use the `_raw` view for `field_is_present` checks** on CSV/Parquet data, so missing columns are truly absent
3. Other checks (type, required, unique, constraints) continue using the unioned table as before

## Files changed
- `duckdb_connection.py`: Create `{model}_raw` view in `create_view_with_schema_union()`
- `data_contract_checks.py`: Pass `use_raw_model=True` for `field_is_present` on csv/parquet
- `tests/test_test_schema_evolution.py`: Update tests to expect correct behavior

All 10 schema evolution tests pass.

Closes #1065


bfgoh:
@hieusats thank you so much for working on this. I need this for my project!

datacontract/engines/data_contract_checks.py
```diff
@@ -192,11 +197,17 @@ def to_schema_name(schema_object: SchemaObject, server_type: str) -> str:
     return schema_object.name
 
 
-def check_property_is_present(model_name, field_name, quoting_config: QuotingConfig = QuotingConfig()) -> Check:
+def check_property_is_present(
+    model_name, field_name, quoting_config: QuotingConfig = QuotingConfig(), use_raw_model: bool = False
```
jschoedl, line 201: To me, `use_raw_model` is not quite self-explanatory - I need to read the code to understand what it does. What about adding a parameter `view_name: str = None` instead, which gets set to `model_name` on `None`?

jschoedl (changes requested):
Thanks! One minor note, and please add a CHANGELOG entry noting the breaking change.

datacontract/engines/soda/connections/duckdb_connection.py
```diff
@@ -96,6 +96,15 @@ def create_view_with_schema_union(con, schema_obj: SchemaObject, model_path: str
         create_empty_table = f"""CREATE TABLE "{model_name}" ({", ".join(columns_def)});"""
         con.sql(create_empty_table)
 
+        # Also create a raw view for field_is_present checks, so missing columns
+        # are actually absent (not filled with NULLs from the unioned table).
+        # See https://github.com/datacontract/datacontract-cli/issues/1065
+        raw_view_name = f"{model_name}_raw"
```
eyupcanakman, line 102: The `field_is_present` bug affects JSON too. Line 73's `read_json_auto(..., columns=...)` fills missing columns with NULL, so the check still passes there. Maybe fix here as well, or as a follow-up?

jschoedl: Good catch. @hieusats You can include this if you like, the required change looks quite similar. But feel free to just resolve this as a follow-up if its getting complicated.

jschoedl:
Hi @hieusats, thank you for this fix! To include it into the our release, I just did a few minor changes to them and merged.

PR #1164, opened 2026-04-18
fix: warn when --output is set but --output-format is missing

### LINKED ISSUE #1156: Fix --output without --output-format silently writing no file
## Reproduction

```bash
datacontract test --output results.xml datacontract.yaml
```

No file is written. No error. No warning. Exit code reflects only test results.

## Expected

1. If `--output-format` is supplied, use it
2. If omitted, infer from extension: `.json` → json, `.xml` → junit
3. If extension or output format is unknown, print error:
   ```
   Error: Cannot infer output format from extension '.txt'. Please specify --output-format (json or junit).
   ```
4. List accepted values in `--output-format` help text

## Files

`datacontract/output/test_results_writer.py` — `write_test_result()` silently skips when `output_format is None` (line 23-29). Apply fix to both `test` and `ci` commands.

barry0451:
## Summary
When `--output <path>` is provided to `lint` or `test` commands without `--output-format`, the CLI silently writes nothing. A yellow warning is now printed guiding the user to also specify `--output-format`.

## Changes
- **File:** `datacontract/output/test_results_writer.py`
- Added a warning branch: if `output_path` is set but `output_format` is `None`, prints a clear message indicating which flag is also needed
- No behavior change for valid invocations

## Testing
- Existing tests pass

Fixes #1156.

### DIFF
--- a/datacontract/output/test_results_writer.py
+++ b/datacontract/output/test_results_writer.py
@@ -22,6 +22,11 @@ def write_test_result(
 ):
     if output_format is not None and not output_path:
         console.print(f"No output path specified for {output_format.value} test results. Skip writing test results.")
+    elif output_path is not None and output_format is None:
+        console.print(
+            "[yellow]Warning:[/yellow] Output path specified but --output-format is missing. "
+            "No test results file will be written. Supported formats: json, junit."
+        )
     elif output_format == OutputFormat.json:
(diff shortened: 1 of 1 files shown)


jochenchrist:
Could we alternatively use a reasonable default output format?

PR #1159, opened 2026-04-15
fix: infer test output format from file extension

dallylee:
## Issue
Fixes #1156.

## Claim
`datacontract test` and `datacontract ci` infer the output format from `--output` when `--output-format` is omitted, and fail clearly for unknown extensions.

## Repro
- Before: `datacontract test --output results.xml datacontract.yaml` wrote nothing.
- After: `.json` writes JSON, `.xml` writes JUnit, unknown extensions exit with a clear error.

## Tests
- `PYTHONPATH=. python3.12 -m pytest tests/test_test_output_json.py tests/test_test_output_junit.py tests/test_cli.py tests/test_ci_output.py -q`
- Result: `34 passed, 8 warnings`

## Why safe
- Explicit `--output-format` still wins.
- Only `.json` and `.xml` are inferred.
- Unknown extensions now fail fast.
- Shared writer path also covers `lint`.

## Hostile self-review
- Strongest wrong argument: this broadens CLI behavior beyond the issue.
  - Answer: the change is centralized and only affects output inference when a file path is already present.
- Strongest too-broad argument: `lint` is included too.
  - Answer: it shares the same writer, so the fix is naturally consistent.
- Strongest proof gap: no end-to-end coverage for `ci`.
  - Answer: `tests/test_ci_output.py` patches `DataContract` and asserts inferred output writes the file.
- Most likely maintainer objection: unknown extensions should not be silently guessed.
  - Answer: they aren't, the command exits with a clear error.

## Evidence
- `evidence/fix-output-without-output-format/`

PR #1173, opened 2026-04-20
fix(output): infer output format from file extension when --output-format omitted

### LINKED ISSUE #1156: Fix --output without --output-format silently writing no file

armorbreak001:
## Problem

Running `datacontract test --output results.xml datacontract.yaml` silently writes no file, produces no error, and gives no warning. The exit code only reflects test results, so the user has no indication that anything went wrong.

This happens because `write_test_result()` in `test_results_writer.py` skips writing when `output_format is None`, and when `--output-format` is not explicitly provided, it defaults to `None`.

## Fix

When `--output` is specified but `--output-format` is omitted, infer the format from the file extension:

- `.json` → `json`
- `.xml` → `junit`

If the extension cannot be mapped (e.g., `.txt`), print a clear error message listing supported formats and exit with code 1.

The inference logic is encapsulated in a small `_infer_output_format()` helper that maps extensions to `OutputFormat` enum values. This keeps the main function clean and makes it easy to extend with new formats later.

## Changes

- **`datacontract/output/test_results_writer.py`**: Added `_infer_output_format()` helper (5 lines) and an early inference block in `write_test_result()` (11 lines) that runs before the existing format dispatch logic. Total: +16 lines, no modifications to existing behavior when `--output-format` is explicitly provided.

Fixes #1156

### DIFF
--- a/datacontract/output/test_results_writer.py
+++ b/datacontract/output/test_results_writer.py
@@ -13,6 +13,13 @@ from datacontract.output.junit_test_results import write_junit_test_results
 from datacontract.output.output_format import OutputFormat
 
 
+def _infer_output_format(output_path: Path) -> Optional[OutputFormat]:
+    """Infer output format from file extension."""
+    ext = output_path.suffix.lower()
+    mapping = {".json": OutputFormat.json, ".xml": OutputFormat.junit}
+    return mapping.get(ext)
+
+
 def write_test_result(
     run: Run,
     console: Console,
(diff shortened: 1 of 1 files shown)


armorbreak001:

jschoedl:
Closing in favour of #1159. Please avoid creating PRs if another one already exists for this issue, except when it is stale for a longer time.

PR #1182, opened 2026-04-20
feat: attach failing row samples to quality check output

### LINKED ISSUE #1180: Quality-check failure output doesn't include a sample failing value
## What happens

Run `test` against a table with a row that violates a quality check.
Example: a contract declares `order_total: {minimum: 0}`, and one row in the table has `order_total = -5.00`.

The CLI correctly reports the check as failed, but the output doesn't say *which row* or *which value* caused the failure:

```json
{
  "check": "order_total minimum: 0",
  "result": "failed",
  "reason": "Value: 1 Fail: {'greaterThan': 0.0, 'lessThan': 0.0}"
}
```

Here `Value: 1` means "1 row failed" — but nothing in the output names the offending row (`order_id='O-BAD123'`) or the failing value (`-5.00`).

kiranannadatha8:
Closes #1180.

## Summary
- When a field quality check fails, we now attach up to 5 sample failing rows to the `Check`, so the offending values show up directly in the JSON / JUnit output and inline in the check's `reason` — no more re-running SQL by hand to figure out what broke.
- Works by injecting `samples limit: N` into the sodacl we already generate, then re-executing Soda's own `failing_sql` query via its connection and attaching the rows.
- Best-effort: if sample capture fails for any reason (non-SQL backend, connection blip, etc.) we log a warning and the check is otherwise untouched.

## Test plan
- [x] New integration test `tests/test_test_quality_samples.py` — postgres container, bad row `{order_id: O-BAD123, order_total: -5}` captured in `failed_rows_samples` and surfaced inline in `reason`.
- [x] Ran existing `test_test_postgres.py` and `test_test_quality.py` — no regressions.
- [x] `ruff check` / `ruff format` clean.

### DIFF
--- a/datacontract/engines/soda/check_soda_execute.py
+++ b/datacontract/engines/soda/check_soda_execute.py
@@ -1,6 +1,11 @@
+import decimal
 import logging
 import typing
 import uuid
+from datetime import date, datetime, time
+from typing import Any
+
+import yaml
 
 from datacontract.engines.soda.connections.athena import to_athena_soda_configuration
 from datacontract.engines.soda.connections.oracle import initialize_client_and_create_soda_configuration
@@ -23,6 +28,8 @@ from datacontract.engines.soda.connections.sqlserver import to_sqlserver_soda_co
 from datacontract.engines.soda.connections.trino import to_trino_soda_configuration
 from datacontract.export.sodacl_exporter import to_sodacl_yaml
 from datacontract.model.run import Check, Log, ResultEnum, Run
+
+FAILED_ROWS_SAMPLES_LIMIT = 5
 
 
 def check_soda_execute(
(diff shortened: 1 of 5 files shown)


jschoedl (changes requested):
Thank you for the PR! To me, this approach reads a bit hacky, as we're relying on Soda internals... Would it also be possible to use a `soda.sampler` for this, or doesn't this work out here fore some reason? https://docs.soda.io/soda-documentation/soda-v3/run-a-scan/failed-row-samples#configure-a-python-custom-sampler

jschoedl:
Hi @kiranannadatha8, thank you for the PR!

We have plans to move away from soda altogether soon. After discussing this internally, we decided we do not want to integrate a Soda-based feature of this scope by now anymore. Sorry for the misleading issue!

PR #1183, opened 2026-04-21
fix(duckdb): use explicit contract schema columns for CSV/Parquet reads (#1065)

### LINKED ISSUE #1065: Check `field_is_present` always passes for CSV/Parquet files

barry0451:
## Summary
Fixes #1065 — `field_is_present` check always passes for CSV/Parquet files because the current implementation compares row counts instead of checking actual column overlap. Uses explicit SELECT of contract schema columns so missing data columns = NULL, allowing `field_is_present` to catch them.

## Changes
- Modified `create_view_with_schema_union` in `datacontract/engines/soda/connections/duckdb_connection.py`
- Previously used INTERSECT to find columns in both contract and data (missing contract columns were silently ignored)
- Now explicitly SELECTs contract schema columns by name, so missing data columns become NULL

## Testing
- All existing tests pass


jschoedl:
Closing in favour of #1163. Please avoid creating PRs if another one already exists for this issue, except when it is stale for a longer time.

PR #1185, opened 2026-04-21
fix(soda): detect missing columns in CSV/Parquet files via field_is_present

### LINKED ISSUE #1065: Check `field_is_present` always passes for CSV/Parquet files

armorbreak001:
## Summary

Fixes #1065 — **`field_is_present` always passes for CSV/Parquet files even when the column is missing from the data.**

## Root Cause

In `create_view_with_schema_union()`, when reading CSV/Parquet files via DuckDB, all columns from the datacontract schema are created in the table (even if they don't exist in the actual data file). Missing columns are filled with NULL values. Soda's `when required column missing` check only verifies column existence in the table — so it always passes, regardless of whether the column was actually in the source file.

## Fix

**Two coordinated changes:**

1. **`duckdb_connection.py`**: After inserting data into the schema-based table, drop any columns that don't exist in the source data file. This ensures Soda's `field_is_present` check correctly detects missing columns.

2. **`check_soda_execute.py`**: Before running Soda scans, pre-filter constraint/quality checks (e.g., `field_required`, `field_minimum`) that reference dropped columns. These are marked as **failed** with reason `"Column X not found in data file"` and excluded from the SodaCL YAML to prevent SQL errors. `field_is_present` checks are intentionally left in so Soda can evaluate them normally (and correctly report them as failed).

## Test Updates

Updated 2 tests to reflect the corrected behavior:
- `test_csv_optional_field_missing_from_old_data` — now expects `field_is_present` failure for missing optional field
- `test_parquet_optional_field_missing_from_old_data` — same for Parquet

All 10 schema evolution tests pass.

### DIFF
--- a/datacontract/engines/soda/check_soda_execute.py
+++ b/datacontract/engines/soda/check_soda_execute.py
@@ -163,18 +163,23 @@ def check_soda_execute(
                 if schema_name != "all" and model_name != schema_name:
                     continue
                 try:
-                    cols = con.sql(f"SELECT column_name FROM information_schema.columns WHERE table_name = '{model_name}'").fetchall()
+                    cols = con.sql(
+                        f"SELECT column_name FROM information_schema.columns WHERE table_name = '{model_name}'"
+                    ).fetchall()
                     table_columns[model_name] = {row[0] for row in cols}
                 except Exception:
                     pass
             if table_columns:
                 filtered_checks = []
                 for check in run.checks:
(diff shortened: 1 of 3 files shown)


armorbreak001:

jschoedl:
Closing in favour of #1163. Please avoid creating PRs if another one already exists for this issue, except when it is stale for a longer time.

PR #1192, opened 2026-04-24
fix(spark): emit native SQL type names as physicalType (#1048)

### LINKED ISSUE #1048: Issue with Databricks Type Export and Validation in CLI Version 0.11.x
We have encountered an issue with the export and validation of Databricks physical types when upgrading from version 0.10.x to 0.11.x. The types that are written into the data contract YAML format have changed, and this change affects the validation process.

In version 0.11.x, the following changes in type representation were noted:
1. string is now represented as StringType()
2. integer is now represented as IntegerType()
3. etc.
 
While validation checks confirm the presence of fields, they incorrectly report their types as None, **without actual validation**:
 
Validation Results:
│ passed │ Check that field 'string_test_1' is present
│ passed │ Check that field string_test_1 has type **None**
│ passed │ Check that field 'bool_test' is present
│ passed │ Check that field bool_test has type **None**
│ passed │ Check that field 'date_test_1' is present
│ passed │ Check that field date_test_1 has type **None** 
│ passed │ Check that field 'num_test_1' is present
│ passed │ Check that field num_test_1 has type **None**

**Expected Behavior:**
**Databricks datatypes should be exported similar to how 0.10.x. OR: Validation should recognize the updated types and validate the fields accordingly rather than defaulting to None.**
 
**Attachments**:
Detailed testing results: [datacontract-databricks-datatypes-export-and-test.xlsx](https://github.com/user-attachments/files/25290253/datacontract-databricks-datatypes-export-and-test.xlsx)

Code Snippet for Data Contract Generation:
from datacontract.data_contract import DataContract
data_contract_specification = DataContract().import_from_source("spark", 'abc.def.table')
data_contract = DataContract(data_contract=data_contract_specification, spark=spark)
contract_yaml = yaml.safe_load(data_contract.export("odcs").replace("\xa0", " "))

jschoedl:
## Summary

Fixes #1048. Since v0.11.0 the Spark importer has emitted Python's `__str__` of the Spark `DataType` object as `physicalType` — `StringType()`, `IntegerType()`, `DecimalType(10,2)`, `ArrayType(StringType(), True)`, etc. These strings aren't recognised by any of the SQL type converters, so they fell through to `return None`, and `datacontract test` then built SodaCL type checks with `expected_type=None` — which SodaCL silently marks as passed. Every affected contract got green "Check that field X has type None" output with no actual validation.

**Why this hid for 5 months**: not missing tests — the importer was covered by an equivalence test that compared its output against `users_datacontract_*.yml` fixtures. But those fixtures themselves contained the buggy `StringType()` strings, so the equivalence held and the test passed. The fixtures were the source of truth for the bug. They've been refreshed to native SQL names, and a new integration test now asserts the round-trip behaviour we actually care about (importer output → real SQL type, not None).

Fix:

1. **Root cause** — `spark_importer.py` now uses `dataType.simpleString()` instead of `str(dataType)`, producing native Spark/Databricks SQL names: `string`, `int`, `bigint`, `decimal(10,2)`, `array<string>`, `struct<number:int,street:string>`, `map<string,int>`.
2. **Defense in depth** — `check_property_type` refuses to build a check with `expected_type=None`, and the engine emits a type check whenever `physicalType` OR `logicalType` is set. When a dialect can't map an explicit `physicalType`, the type is passed through verbatim so SodaCL compares it against the real DB column — a typo or wrong physicalType fails loudly instead of silently skipping.
3. **Warnings on unmapped types** — every SQL type converter logs a warning at its terminal fallthrough. The warning detects legacy Spark-repr physicalTypes (`StringType()`, etc.) and points users to re-import.
4. **Coverage for simpleString types** — `binary`, `tinyint`, `smallint` now route to real SQL types in the `databricks`, `dataframe`, `snowflake`, `postgres`, and `mysql` converters.
5. **Also fixed while in the area**:
   - `_get_base_type` now strips `<...>` parameterisation (needed for `array<string>`, `struct<...>`, `map<...>`).
   - Decimal precision/scale parsed directly from `decimal(p,s)` strings.
   - Removed a stray `from venv import logger` in `data_contract_checks.py` (was importing the stdlib `venv` module's logger).
   - **BigQuery**: `logicalType: double` was mapping to `BIGNUMERIC` (a high-precision fixed-point decimal) — corrected to `FLOAT64`.
   - **Trino**: `binary` was falling through to the unmapped warning — now maps to `varbinary` alongside `bytes`.
   - **DuckDB**: `tinyint` and `smallint` were falling through — now map to native `TINYINT` / `SMALLINT`.
   - **Oracle**: `boolean` was mapping to bare `CHAR` — tightened to `CHAR(1)`.

**Breaking**: contracts produced by v0.11.0–v0.12.1 with Spark-repr physicalTypes (`StringType()`, etc.) must be re-imported. There is no compat shim for the legacy form, but the warning emitted at SQL conversion time tells users to re-import.

Known limitation (pre-existing, out of scope): Spark `MapType` has no native representation in ODCS, so map-typed columns still don't produce a usable Databricks SQL type.

## Test plan

- [x] `pytest tests/test_import_spark.py` — 6/6 pass
- [x] Dialect cross-check: BigQuery / Trino / DuckDB / Oracle — 100+ tests pass
- [x] Full non-PySpark suite — 620 passed, 13 skipped, 0 failures
- [x] `ruff check .` and `ruff format .` — clean

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - `changelog` command help text now advertises `(url or path)` for V1/V2 arguments, clarifying that HTTP/HTTPS URLs are accepted (#1162)
 - **breaking:** `test` command now exits non-zero when a server is specified, but soda-core fails to connect or authenticate (#1181)
 - correct swapped `check_type` labels  `model_qualty_sql` and `field_quality_sql` (#1187)
+- `import spark` now emits native Spark SQL type names (`string`, `int`, `decimal(10,2)`, `array<string>`, …) as `physicalType` instead of Python repr (`StringType()`, `IntegerType()`, …); type checks in `test` no longer silently pass with `has type None` (#1048)
 
 ## [0.12.1] - 2026-04-21
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,7 +11,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - `changelog` command help text now advertises `(url or path)` for V1/V2 arguments, clarifying that HTTP/HTTPS URLs are accepted (#1162)
 - **breaking:** `test` command now exits non-zero when a server is specified, but soda-core fails to connect or authenticate (#1181)
 - correct swapped `check_type` labels  `model_qualty_sql` and `field_quality_sql` (#1187)
-- `import spark` now emits native Spark SQL type names (`string`, `int`, `decimal(10,2)`, `array<string>`, …) as `physicalType` instead of Python repr (`StringType()`, `IntegerType()`, …); type checks in `test` no longer silently pass with `has type None` (#1048)
+- `import spark` now emits a native Spark SQL physicalType (e.g. `string`) instead of Python repr (e.g. `StringType()`). Contracts imported using Sparkk in v0.11.0–v0.12.1 did not perform type checks and must be re-imported. (#1048)
 
 ## [0.12.1] - 2026-04-21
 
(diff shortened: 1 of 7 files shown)


jschoedl:
Seems like the most recent release of `mysql-connecter-python` is incomplete, making the checks fail.

PR #1199, opened 2026-04-29
Fix/trino http scheme

the-romiro:
### Summary

This PR add `http_scheme` propertie for fixes an issue where the `http_scheme` defined in `data-contract.yml` was ignored during the conversion from DCS to ODCS.

Trino connections in soda-core are `https`, even when `http` was explicitly configured. This breaks common local and non-SSL setups.

### Root Cause

Although `http_scheme` is a valid field in the Data Contract specification, it was not being propagated during the internal model transformation (DCS → ODCS). Consequently, the value never reached the connection layer in soda-core.

### Changes

* Added `http_scheme` to the `Server` model in both:

  * DCS (Data Contract Specification)
  * ODCS (Operational Data Contract Specification)
* Fixed the mapping logic to ensure the field is preserved during DCS → ODCS conversion
* Ensured the value is correctly passed to the soda-core connection layer

### Why this matters

The Trino connector in soda-core supports `http_scheme="http"` as the official way to connect to instances without SSL (e.g., local environments), as seen in:

* [`trino_data_source_connection.py`](https://github.com/sodadata/soda-core/blob/main/soda-trino/src/soda_trino/common/data_sources/trino_data_source_connection.py#L24)

Without this fix, users cannot connect to:

* Local Trino instances
* Test environments without HTTPS
* Custom deployments without SSL/Auth

### Example

```yaml
server:
  host: localhost
  port: 8080
  http_scheme: http
```

Before this fix:

* `http_scheme` was ignored → connection attempted via `https`

After this fix:

* `http_scheme` is respected → connection uses `http` as expected

### Impact

* Enables local development workflows with Trino

- [x] Code formatted (`uv run ruff format`)
- ~~[ ] README.md updated (if relevant)~~

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## Unreleased
 
+### Added
+- Add `http_scheme` support for Trino connections, allowing HTTP or HTTPS to be configured via server config or `model_extra`
+- Propagate `http_scheme` field when importing DCS servers to ODCS format
+
 ### Fixed
 - `changelog` command help text now advertises `(url or path)` for V1/V2 arguments, clarifying that HTTP/HTTPS URLs are accepted (#1162)
 - **breaking:** `test` command now exits non-zero when a server is specified, but soda-core fails to connect or authenticate (#1181)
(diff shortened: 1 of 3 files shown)


the-romiro:
If we specify the HTTP scheme in the host, the connection will be OK.

See: [example](https://github.com/datacontract/datacontract-cli/blob/main/tests/fixtures/trino/datacontract.yaml#L10)

```yaml
server:
  type: trino
  host: http://localhost
  port: 8080
```

PR #1211, opened 2026-05-04
feat(dependencies): add databricks-runtime extra

chifu1234:
This new optional dependency group includes all the soda-core-spark, databricks-sql-connector, and databricks-sdk dependencies from the databricks extra, but excludes pyspark (provided by the Databricks Runtime environment).

Use `pip install "datacontract-cli[databricks-runtime]"` when running inside Databricks to avoid dependency conflicts.

Updated CHANGELOG.md and README.md with the new extra.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## Unreleased
 
+### Added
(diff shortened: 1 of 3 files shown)


jschoedl:
This should resolve a conflict with pyspark, right? But `soda-core-spark-df==3.5.6` (which you included) pulls in `pyspark==3.5.8` anyways...


Can you name the exact conflict you are experiencing?

jochenchrist:
With v1.0.0 the dependencies have changed from Soda to ibis, which also might make this feature easier. 

I think this extra still makes sense to better support Databricks environments.
Could you please create a new / updated PR?

simonharrer:
Great idea. I needed the change in, so I took over, see https://github.com/datacontract/datacontract-cli/commit/b1d08e979cd9703809dab135a93e7e8d0dbadeef and added you as a co-author of that work that is now based on ibis. I hope this is fine for you.

PR #1212, opened 2026-05-05
feat(quality): support {object} placeholder in SQL quality queries

### LINKED ISSUE #676: Add support for placeholders in ODCS Data Contracts
The ODCS format defines a different syntax and naming for placeholders in data quality definitions. See https://bitol-io.github.io/open-data-contract-standard/latest/#valid-properties-for-operator

To fully support ODCS, datacontract-cli should support the `${}` syntax and the additional placeholders `${object}` and `${property}`.

simonharrer:
## Summary

- Adds `{object}` / `${object}` substitution to `prepare_query` in `data_contract_checks.py` as an alias for `{model}`/`{table}`. Picks up the same dialect-aware quoting.
- Brings the CLI in line with the [ODCS spec](https://bitol-io.github.io/open-data-contract-standard/latest/#valid-properties-for-operator), which documents `{object}` and `{property}` as the canonical SQL-quality placeholders.
- Closes #676. Also makes `datacontract-editor`'s existing `Use {object} and {property} as placeholders` hint correct end-to-end.

### Why now

Found while debugging why `{object}` leaks unreplaced into quality SQL. `{object}` was never substituted in any released version of the CLI (verified across all 24 historical revisions of `prepare_query`). Asymmetry crept in with #783, which added `{property}` (the ODCS-spec property alias) but missed adding `{object}` alongside `{model}`/`{table}`.

## Test plan

- [x] `pytest tests/test_data_contract_checks.py` — 41/41 pass
- [x] `ruff check` / `ruff format --check` clean
- [x] New tests cover bare `{object}`, `${object}`, and the canonical ODCS example `SELECT COUNT(*) FROM {object} WHERE {property} IS NOT NULL`
- [ ] CI green

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## Unreleased
 
 ### Fixed
+- quality: support `{object}` and `${object}` placeholder in SQL quality queries as the ODCS-spec name for the current schema object (alias for `{model}`/`{table}`) (#676)
 - `changelog` command help text now advertises `(url or path)` for V1/V2 arguments, clarifying that HTTP/HTTPS URLs are accepted (#1162)
 - **breaking:** `test` command now exits non-zero when a server is specified, but soda-core fails to connect or authenticate (#1181)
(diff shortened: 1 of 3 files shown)


jochenchrist (approved):
LGTM

PR #1213, opened 2026-05-05
chore!: drop dbt-core dependency, parse manifest.json directly

simonharrer:
## Summary

- The `dbt` importer only used `dbt-core` to parse `manifest.json` and call `ModelNode.infer_primary_key`. Both are easily replaceable: the manifest is plain JSON, and `infer_primary_key` is ~30 lines of test-graph traversal.
- Rewrote `datacontract/imports/dbt_importer.py` to read the manifest as a `dict` and reimplemented `_infer_primary_keys` (same precedence: PK constraint → unique+not_null → enabled-unique → disabled-unique).
- Enforce minimum manifest schema **v9** (dbt 1.5+) with a clear error for older manifests. Reads `metadata.dbt_schema_version`.
- Dropped the `[dbt]` extra and `dbt-core` dependency. `datacontract import dbt` now works out of the box with no extra needed.
- Updated the `mysql-connector-python` override comment in `[tool.uv]` (the dbt-core conflict that motivated it is gone).

### Why

`dbt-core` is a heavy dependency (drags in agate, snowplow-tracker, jinja, networkx, protobuf, …) for what amounts to JSON parsing. It also conflicted with `mysql-connector-python` (see [soda-core#2515](https://github.com/sodadata/soda-core/issues/2515)), motivating the override workaround. With dbt-core gone, the protobuf conflict no longer applies.

### Behavior

Identical ODCS output for all existing dbt manifest fixtures. Smoke-tested by blocking `dbt`/`dbt_common` modules from import — the importer still parses the jaffle fixtures correctly.

### Breaking change

Users who installed `datacontract-cli[dbt]` should switch to plain `datacontract-cli`. Manifests below schema v9 (dbt < 1.5, May 2023) are no longer accepted; regenerate with a current dbt version.

## Test plan

- [x] `pytest tests/test_import_dbt.py` — 8/8 pass; expected ODCS YAML unchanged against jaffle duckdb / bigquery / empty-columns fixtures
- [x] `pytest tests/test_export_dbt_*.py` — 11/11 pass (no collateral breakage)
- [x] `ruff check` / `ruff format --check` clean
- [x] Smoke: import works with `dbt` and `dbt_common` blocked via `sys.meta_path`
- [ ] CI green

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -169,9 +169,6 @@ requires = ["setuptools", "wheel"]
 build-backend = "setuptools.build_meta"
 
 [tool.uv]
-# soda-core-mysql pins mysql-connector-python==8.0.30 (from 2022). Newer
-# mysql-connector-python (9.x) works fine with soda. See:
-# https://github.com/sodadata/soda-core/issues/2515
 override-dependencies = [
     "mysql-connector-python>=8.0.30,<9.7.0",
 ]
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -110,10 +110,6 @@ impala = [
   "soda-core-impala>=3.3.20,<3.6.0"
 ]
 
-dbt = [
-  "dbt-core>=1.8.0"
-]
-
 dbml = [
   "pydbml>=1.1.1"
 ]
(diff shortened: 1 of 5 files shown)


pyproject.toml
```diff
@@ -169,9 +165,9 @@ requires = ["setuptools", "wheel"]
 build-backend = "setuptools.build_meta"
 
 [tool.uv]
-# soda-core-mysql pins mysql-connector-python==8.0.30 (from 2022) which requires
-# protobuf<=3.20.1, conflicting with dbt-core. Newer mysql-connector-python (9.x)
-# works fine with soda and resolves the conflict. See: https://github.com/sodadata/soda-core/issues/2515
+# soda-core-mysql pins mysql-connector-python==8.0.30 (from 2022). Newer
```
jochenchrist: Can we get rid of this comment?

jochenchrist: @copilot

PR #1238, opened 2026-05-18
fix(sql-type): canonicalize MySQL numeric -> decimal

jschoedl:
## Summary

Mirror of #1237, this time for MySQL. `convert_type_to_mysql` emitted `numeric(...)` for ODCS `type: numeric`, but MySQL treats `DECIMAL` and `NUMERIC` as exact synonyms and `information_schema.columns.data_type` always reports `decimal`. As a result, `datacontract test` failed the column-type check for any contract that used `type: numeric` against MySQL:

```
Check that field amount has type numeric:
  Type Mismatch, Expected Type: numeric; Actual Type: decimal
```

This patch maps both ODCS `decimal` and `numeric` base types to `decimal` in `convert_type_to_mysql`. The emitted SQL is functionally identical in MySQL; the type check now matches what the database actually reports.

## Test plan

- [x] Live verification against MySQL 8 via the existing `test_test_mysql.py` testcontainers tests. Added `field_four NUMERIC(10,2)` to the MySQL fixtures (`datacontract.yaml`, `odcs.yaml`, `data.sql`) so the test exercises the `numeric` -> `decimal` mapping. Both `test_test_mysql` (DCS) and `test_test_mysql_odcs` (ODCS) pass.
- [x] `pytest tests/test_sql_type_converter_physicaltype.py tests/test_export_sql.py` — 50 passed, including a new `test_numeric_10_2_mysql` and `test_logical_numeric_mysql`.
- [x] `ruff check` + `ruff format --check` clean.

Without the converter fix, the new fixture column makes the testcontainers test fail with the exact `Type Mismatch` error above — that's the live proof the fix works.

## Notes

- Same shape as #1237 (Postgres) but the canonical direction is flipped: Postgres canonicalizes to `numeric`, MySQL canonicalizes to `decimal`.
- SQL Server might have a similar issue and is left out of scope here.


jschoedl:
Folded into #1237. The MySQL commit is now part of the combined Postgres + MySQL decimal/numeric canonicalization PR.

PR #1247, opened 2026-05-20
feat(test): resolve semantic definitions to enrich properties at read time

andredlng:
## Summary

Resolves `authoritativeDefinitions[type=semantics]` on each property at read time so `datacontract test` works against the *lean-YAML* model some editors adopt: a property carrying only a `name`, `physicalName`, and a semantic URL gets its `logicalType` / `description` / `examples` / `businessName` filled from the referenced concept just before checks are generated. The contract on disk is never modified.

Without this, properties produced by the datacontract-editor's "Add from definition" get a presence check but no type check — the schema validation silently degrades because `to_schema_checks` gates the type check on `prop.physicalType or prop.logicalType` being present, and neither is set inline.

## How it works

- `datacontract/integration/entropy_data.py` gains `resolve_semantic_definition(url)`: translates the URL `https://<host>/<vanity>/semantics/<namespace>/<externalId>` to the API path `https://<host>/api/semantics/experimental/namespaces/<namespace>/concepts/<externalId>`, fetches the concept JSON, and caches it per process. Sits alongside the existing `publish_*` helpers and reuses `_get_api_key()`.
- `datacontract/integration/semantic_enrichment.py` walks every property in the in-memory contract and merges the four concept fields above when (and only when) the property doesn't have them inline.
- `create_checks` calls `enrich_in_place(...)` before walking properties. No exporter call sites change; they remain at today's behavior and can opt in in a follow-up.

## Security

The API key is attached **only** when the URL's host matches `*.entropy-data.com` / `entropy-data.com` *or* the configured `ENTROPY_DATA_HOST` / `DATAMESH_MANAGER_HOST` / `DATACONTRACT_MANAGER_HOST`. Any other host (a third-party URL named by the contract author) is queried anonymously. A contract cannot redirect the resolver at an arbitrary host and exfiltrate the user's key. Covered by a unit test.

## Self-hosted deployments

Works out of the box for any standard self-hosted entropy-data instance (the same Spring app, the same URL layout). The user sets `ENTROPY_DATA_HOST=https://entropy.acme.internal` and `ENTROPY_DATA_API_KEY=ed_…` (the same plumbing the existing publish code uses), and contract URLs on that host resolve with the key attached. The existing `--no-ssl-verification` flag on `datacontract test` (and `ci`/`dbt`/`publish`) now threads through `create_checks` → `enrich_in_place` → `resolve_semantic_definition` to the underlying `requests.get`, supporting self-signed certificates.

## Failure behavior

Resolution failures — no network, no key, untrusted host returning non-200, network exception, malformed JSON, URL that doesn't match the expected shape — return `None` and the field stays unfilled. The resolver's per-process cache stores `None` too, so a single test run never makes the same failing request twice. The check set in that case is exactly today's, so the change is additive and never regresses behavior.

## Tests

`tests/test_semantic_enrichment.py` (23 tests) covers:
- URL translation: valid shapes, query/fragment-stripping, rejected shapes (empty, bad scheme, missing segments).
- Host-trust policy: `*.entropy-data.com` accepted, arbitrary hosts rejected, configured host accepted.
- Resolver: 200 returns concept; key sent to trusted host; key withheld from untrusted host; 200 caches; 404 caches negative; network error returns `None`; malformed URL short-circuits without HTTP.
- Enrichment: missing fields filled; inline fields not overwritten; no-op without a semantic ref; safe on contracts with no schema.
- End-to-end: `create_checks` emits a `field_type` check for a previously-bare semantic property.
- SSL verification: `ssl_verification=True` (default) and `ssl_verification=False` both propagate through `create_checks` to `requests.get(verify=…)`.

Existing `tests/test_data_contract_checks.py` (43 tests) is unchanged and still green.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 - `datacontract test` now logs the Data Contract CLI version and whether it ran as a local CLI or through the FastAPI server (including the request URL) as part of the test result logs
+- `datacontract test` now resolves `authoritativeDefinitions[type=semantics]` on each property and fills missing `logicalType` / `description` / `examples` / `businessName` from the concept at run time. A property that only carries a semantic ref (e.g. produced by the datacontract-editor's "Add from definition") now generates presence and type checks instead of silently skipping them. The API key is sent only to trusted hosts (`*.entropy-data.com` and the netloc of `ENTROPY_DATA_HOST` / `DATAMESH_MANAGER_HOST` / `DATACONTRACT_MANAGER_HOST`); other hosts are queried anonymously. Honors the existing `--no-ssl-verification` flag for self-signed certificates in self-hosted deployments. Concepts are cached per-process
 
 ### Fixed
 - Schema checks now resolve each property by its `physicalName` when set (falling back to `name`), matching the existing table-level resolution and the SQL/BigQuery exporters. Previously a property whose logical `name` differed from its physical column (e.g. `name: brand` with `physicalName: BRAND`) failed the presence and type checks even though the physical column existed (#1246)
--- a/datacontract/data_contract.py
+++ b/datacontract/data_contract.py
@@ -147,6 +147,7 @@ class DataContract:
                 self._duckdb_connection,
                 schema_name=self._schema_name,
                 check_categories=self._check_categories,
+                ssl_verification=self._ssl_verification,
             )
 
         except DataContractException as e:
--- a/datacontract/engines/data_contract_checks.py
+++ b/datacontract/engines/data_contract_checks.py
@@ -14,6 +14,7 @@ from open_data_contract_standard.model import (
 )
 
 from datacontract.export.sql_type_converter import convert_to_sql_type
+from datacontract.integration.semantic_enrichment import enrich_in_place
 from datacontract.model.run import Check
 
 logger = logging.getLogger(__name__)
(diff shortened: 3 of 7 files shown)


andredlng:
Closing for now — not required for the dataproduct-builder demo to work visibly. The R6a field-level `physicalName` fix (#1246, already merged) is what lands the demo's green-checkmark payoff. This PR closes a separate, silent correctness gap on top of that, which nobody has been bitten by yet.

**What's deferred together with this PR:**

- The silent type-skip in `to_schema_checks` for properties that carry only `authoritativeDefinitions[type=semantics]` and no inline `logicalType`/`physicalType`. With this closed, a semantic-only property still gets a presence check but no type check — deploying a wrong-typed column would pass `datacontract test` green.
- A reusable scaffold for resolving other `authoritativeDefinitions` types — most obviously `businessDefinition`, which is in the same shape and equally unhandled today.

**Revisit when** somebody actually hits the silent-skip (a wrong-typed column going undetected for a semantic-only property), or when `businessDefinition` resolution becomes needed.

**Starting points if/when we come back:** the implementation here (resolver + walker + host-trust gate + cache + ssl threading; 23 tests; CI fully green) is reusable as-is. The two known follow-ups discussed in this PR's thread are (1) IRI handling — platform needs a public `GET /api/semantics/.../concepts?iri=...` endpoint (the repository method `findAllByOrganizationIdAndIri` already exists internally), then a small fallback branch in `_semantic_url_to_api`; and (2) a refactor to a type-dispatched registry so `semantics` and `businessDefinition` share one scaffold.

PR #1248, opened 2026-05-20
feat: support for relative (percent) metrics in data contract checks.

### LINKED ISSUE #1228: Add support for inbuilt quality metrics with relative thresholds as per ODCS 3.1.
OIDC 3.1 describe quality metrics for both absolute (rows) and relative (percent) thresholds:

https://bitol-io.github.io/open-data-contract-standard/v3.1.0/data-quality/#null-values:~:text=Example%20with%20percent%3A

Image

 I'll be happy to open a PR if you are taking contributions.

HenrikSpiegel:
- [X] Tests pass (`uv run pytest`)
- [X] Code formatted (`uv run ruff check --fix && uv run ruff format`)

https://github.com/datacontract/datacontract-cli/issues/1228


datacontract/engines/data_contract_checks.py
```diff
@@ -922,32 +928,39 @@ def prepare_query(
...
 
 def to_sodacl_threshold(quality: DataQuality) -> str | None:
+    if quality.unit is not None and quality.unit not in ("rows", "percent"):
+        logger.warning(
+            f"Unsupported quality.unit ={quality.unit} in quality check, must be 'rows' or 'percent' or None"
+        )
+        return None
+    threshold_suffix = "%" if quality.unit == "percent" else ""
+
     if quality.mustBe is not None:
```
HenrikSpiegel, line 978: discussion:

I initially implemented this going for minimal changes and ensuring the changes matched the style of the repository.

I would however personally prefer to re-write the function using a operator mapping:
```
arg_to_op = {
  'mustBe' : '=',
  'mustNotBe': '!='
}
for arg, op in arg_to_op.items():
  if val := getattr(arg, quality) is not none:
    return f"{op} {val}{threshold_suffix}"

```

jschoedl: I get your point, but this won't work out for the 'between' cases, right?
The reason for this repetitive style is that it is easier to understand the code with all relevant cases, whereas a mapping as you proposed has more mental overhead to read. Of course this is a trade-off, since copying things like the `{threshold_suffix}` you added can be error-prone as well. But in this case, I'd vote for keeping it as is.

datacontract/engines/data_contract_checks.py
```diff
@@ -558,6 +558,8 @@ def check_property_regex(
 
 def check_row_count(model_name: str, threshold: str, quoting_config: QuotingConfig = QuotingConfig()):
     check_type = "row_count"
+    if "%" in threshold:
+        logger.warning("Row count threshold cannot be specified as a percentage.")
```
jschoedl, line 601: After the warning, we should not create the check (Soda would ignore the percent sign and only use the integer, which is not what the user is intending)

datacontract/engines/data_contract_checks.py
```diff
@@ -922,32 +928,39 @@ def prepare_query(
 
 
 def to_sodacl_threshold(quality: DataQuality) -> str | None:
+    if quality.unit is not None and quality.unit not in ("rows", "percent"):
+        logger.warning(
+            f"Unsupported quality.unit ={quality.unit} in quality check, must be 'rows' or 'percent' or None"
+        )
```
jschoedl:
```suggestion
            f"Unsupported quality.unit '{quality.unit}' in quality check, must be 'rows' (default) or 'percent'"
        )
```

`None` basically means that the value is not present and `'rows'` will be used.

datacontract/engines/data_contract_checks.py
```diff
@@ -922,32 +928,39 @@ def prepare_query(
 
 
 def to_sodacl_threshold(quality: DataQuality) -> str | None:
+    if quality.unit is not None and quality.unit not in ("rows", "percent"):
+        logger.warning(
```
jschoedl:
```suggestion
    if quality.unit is not None and quality.unit.lower() not in ("rows", "percent"):
```

datacontract/engines/data_contract_checks.py
```diff
@@ -922,32 +928,39 @@ def prepare_query(
...
+            f"Unsupported quality.unit ={quality.unit} in quality check, must be 'rows' or 'percent' or None"
+        )
+        return None
+    threshold_suffix = "%" if quality.unit == "percent" else ""
+
```
jschoedl:
```suggestion
    threshold_suffix = "%" if quality.unit.lower() == "percent" else ""

```

datacontract/engines/data_contract_checks.py
jschoedl: This still includes `duplicate_count`. (5 similar locations need to be fixed)

jschoedl (changes requested):
Thank you for the PR! A few notes, and please add test cases for this in the `tests` directory - e.g.:
- test that `percent` really appends `%`
- test that no `%` exists for the default case or `rows`
- test that e.g. `check_property_null_values` with percent produces the correct sodacl check
- test that a percent sign in `check_row_count` behaves correctly

datacontract/engines/data_contract_checks.py
HenrikSpiegel: Completely fair - I'll let it be.

You are right wrt. to the `between` operator that would require handling as a special case.
Since the function pattern is using early return - one would just have to add an if-statement for between after the loop.

datacontract/engines/data_contract_checks.py
HenrikSpiegel: good suggested - implemented while handling Nones

datacontract/engines/data_contract_checks.py
HenrikSpiegel: good suggestion.

datacontract/engines/data_contract_checks.py
HenrikSpiegel: Good catch - fixed in [484640b](https://github.com/datacontract/datacontract-cli/pull/1248/commits/484640b3f3da1ee4398abe5d3e7bd53f81f2aa78)

datacontract/engines/data_contract_checks.py
HenrikSpiegel: fxied

datacontract/engines/data_contract_checks.py
HenrikSpiegel: Added None return

jochenchrist:
Thanks for this, and apologies it sat open long enough to be overtaken by events.

This is now superseded by the v1.0.0 refactor in #1279 ("Replace Soda Core with ibis test engine"), which landed the same feature through a different path:

- **The target file is gone.** `datacontract/engines/data_contract_checks.py` was removed in #1279; the check-building code moved to `datacontract/export/sodacl_check_builder.py`. This PR no longer applies cleanly.
- **Percent thresholds are now supported natively.** `datacontract test` honors ODCS `quality.unit: percent` on the count-of-bad-rows metrics (`nullValues`, `missingValues`, `invalidValues`), comparing the threshold against the failed fraction of the row count. Percent on metrics where a row fraction has no meaning (`rowCount`, `duplicateValues`) logs a warning and falls back to the absolute count, matching the behavior in this PR. See `is_percent_unit()` in `datacontract/engines/checks/create_checks.py`, `threshold_is_percent` in `check_spec.py`, the comparison logic in `ibis/ibis_check_execute.py`, and `tests/test_test_quality_percent_severity.py`.

That resolves #1228, so I'm closing this as already implemented.

One thing not covered by #1279: the SodaCL **export** (`sodacl_check_builder.py`) still emits count-only metrics and doesn't apply a `%` suffix. If emitting percent thresholds in the SodaCL export is still useful to you, a smaller follow-up scoped to that file would be welcome.

PR #1249, opened 2026-05-20
fix: support versioned dbt schemas in `import dbt` commands

willbowditch:
### Problem

Importing a specific version of a dbt model using the `name.vN` convention produced an empty contract with no error:

```bash
# Silent empty output
datacontract import dbt --source target/manifest.json --model mart_orders.v1
datacontract import dbt --source target/manifest.json --model mart_orders.v2
```

In `manifest.json`, dbt always stores the base name in `node["name"]` and the version number separately in `node["version"]`. The filter compared the user-supplied string (`"mart_orders.v1"`) directly against `node["name"]` (`"mart_orders"`), so every node was skipped.

### Fix

Added `_matches_dbt_node_filter()` in `dbt_importer.py` which understands the `name.vN` convention:

- `--model mart_orders` → imports all versions (unchanged behaviour)
- `--model mart_orders.v1` → imports only the v1 node
- `--model mart_orders.v2` → imports only the v2 node

### Tests

Added a versioned manifest fixture (`manifest_versioned_models.json`) and four new tests covering the above cases.

### Checklist

- [x] Tests pass (uv run pytest)
- [x] Code formatted (uv run ruff check --fix && uv run ruff format)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 - Schema checks now resolve each property by its `physicalName` when set (falling back to `name`), matching the existing table-level resolution and the SQL/BigQuery exporters. Previously a property whose logical `name` differed from its physical column (e.g. `name: brand` with `physicalName: BRAND`) failed the presence and type checks even though the physical column existed (#1246)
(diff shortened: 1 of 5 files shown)


jschoedl (approved):
LGTM! I don't think we need that many tests for `_matches_dbt_node_filter`, the other behavioral tests should be enough.

PR #1253, opened 2026-05-22
fix: skip schema type check for varchar(n) and map columns on Databricks (#1219, #1245)

### LINKED ISSUE #1219: Validation still fails for structs containing maps after #1048 fix
## Summary

I retested the fix for #1048 in v0.12.2 and the situation is much better overall.
Most Databricks/Spark types now validate correctly.

I found one remaining case: when a struct contains a nested map field, validation builds the expected type with `None` for that nested map and then fails with a type mismatch.

## Related issue

Follow-up to #1048.

## Current behavior

For a field with this exported physical type:

`struct<complex:string,info:struct<age:int,addressbook:map<string,string>>>`

validation fails with:

```text
Type Mismatch
Expected Type:
STRUCT<complex:STRING,info:STRUCT<age:INT,addressbook:None>>
Actual Type:
struct<complex:string,info:struct<age:int,addressbook:map<string,string>>>
```

So the nested `map<string,string>` appears to be converted to `None` during expected-type creation.

## Expected behavior

One of these should happen:

1. The nested map type is preserved and validated correctly inside the struct.
2. If nested maps are not supported for validation, they should be handled explicitly with a warning rather than being converted to `None` and failing with a type mismatch.

## Reproduction

The issue is reproducible with a contract that contains a property like this:

```yaml
  - name: struct_with_map_test
    logicalType: object
    physicalType: struct<complex:string,info:struct<age:int,addressbook:map<string,string>>>
    properties:
      - name: complex
        logicalType: string
        physicalType: string
      - name: info
        logicalType: object
        physicalType: struct<age:int,addressbook:map<string,string>>
        properties:
          - name: age
            logicalType: integer
            physicalType: int
          - name: addressbook
            logicalType: object
            physicalType: map<string,string>
```

This is how I generated and tested the contract:

```python
import yaml
from datacontract.data_contract import DataContract

data_contract_specification = DataContract().import_from_source("spark", tbl_str)
data_contract = DataContract(data_contract=data_contract_specification, spark=spark)
contract_dict = yaml.safe_load(data_contract.export("odcs"))

with open(contract_path_test, "w", encoding="utf-8") as f:
  yaml.dump(contract_dict, f)

data_contract = DataContract(
  data_contract_file=contract_path_test,
  spark=spark,
  server=test_server_name,
)

run = data_contract.test()
```

And this is the failing output:

```text
data contract is invalid, found the following errors: Type Mismatch
Expected Type:
STRUCT<complex:STRING,info:STRUCT<age:INT,addressbook:None>>
Actual Type:
struct<complex:string,info:struct<age:int,addressbook:map<string,string>>>
```

## Environment

- datacontract-cli: 0.12.2
- datacontract-specification: 1.2.3
- open-data-contract-standard: 3.1.2
- soda-core: 3.5.6
- soda-core-spark: 3.5.6
- soda-core-spark-dfpydantic: not installed
- pydantic_core: 2.27.2
- pyspark: 3.5.8
- fastjsonschema: 2.21.1
- jsonschema: 4.23.0
- jsonschema-specifications: 2023.7.1
- PyYAML: 6.0.2
- numpy: 2.1.3

## Notes

This appears to be the only remaining case I found while testing the #1048 fix.

### LINKED ISSUE #1245: Schema validation always fails for varchar(n) columns on Databricks Serverless (DBR 14+)
Schema type validation for `varchar(n)` columns always fails when validating against a Databricks Serverless cluster (Databricks Runtime 14+), regardless of whether the data contract specifies `physicalType: string` or `physicalType: varchar(n)`.

### Steps to reproduce

1. Create a Databricks Delta table with a `varchar(30)` column (on DBR 14+ / Serverless)
2. Define a data contract with `physicalType: varchar(30)` for that column
3. Run `DataContract(..., spark=spark, server="dbx_prod").test()` via Databricks Connect (serverless)
4. Observe the schema check failure

### Expected behavior

`varchar(n)` columns should pass schema validation when the contract specifies `physicalType: varchar(n)`.

### Actual behavior

```
Type Mismatch, Expected Type: STRING; Actual Type: varchar(30)
```

The check also fails if `physicalType` is set to string in the contract.

### Root cause

There are two interacting issues.

**1. `convert_to_databricks()` in `sql_type_converter.py` unconditionally strips the length from `varchar(n)`:**

```python
# sql_type_converter.py
if base_type in ["string", "varchar", "text"]:
    return "STRING"   # ← length is lost; varchar(30) becomes STRING
```

`_get_base_type("varchar(30)")` returns `"varchar"`, which maps to `"STRING"`. The generated SodaCL check therefore always uses `STRING` as the expected type, regardless of whether the contract says `string` or `varchar(30)`.

**2. `soda-core-spark-df`'s `SCHEMA_CHECK_TYPES_MAPPING` does not cover `varchar(n)` with a length suffix:**

```python
# spark_df_data_source.py (soda-core-spark-df 3.5.4 / 3.5.6)
SCHEMA_CHECK_TYPES_MAPPING = {
    "string": ["character varying", "varchar", "text"],  # no "varchar(30)"!
    ...
}
```

`is_same_type_in_schema_check("STRING", "varchar(30)")` returns `False` because `"varchar(30)"` is not in the alias list for `"string"`.

**Why this regressed:** On older Databricks Runtimes (≤13 / classic clusters), `INFORMATION_SCHEMA.COLUMNS.data_type` returned `string` for `varchar(n)` columns. Databricks Serverless (DBR 14+) now correctly preserves the full `varchar(30)` type. The old workaround of mapping everything to `STRING` no longer holds.

### Environment

- `datacontract-cli`: 0.12.2
- Databricks Serverless (PySpark 4.0.0 / DBR 14+)
- Databricks Connect 17.3
- soda-core-spark-df: 3.5.4 / 3.5.6 (both affected).

### Suggested fix

In `convert_to_databricks()`, preserve the length for `varchar(n)` instead of normalizing to `STRING`:

```python
def convert_to_databricks(field):
    ...
      if base_type in ["string", "text"]:
          return "STRING"
      if base_type == "varchar":
          # Preserve varchar(n) with length so Soda exact-matches what DBR 14+ reports;
          # plain varchar (no length) falls back to STRING
          return _attach_params_if_present("varchar", field) or "STRING"
```

IchEssBlumen:
## Problem

`soda-core-spark-df` ≤3.5.6 (unmaintained) cannot match certain column types correctly on Databricks with PySpark 4.0+

- **`varchar(n)`** – PySpark 4.0+ preserves `VarcharType` as a distinct type, which is not in `SCHEMA_CHECK_TYPES_MAPPING`, causing every `varchar(n)` column to fail the type check
- **`map` nested inside a struct** – `convert_to_databricks()` returns `None` for nested maps, producing an invalid expected type like `STRUCT<...,field:None>`
- **struct columns containing `varchar(n)` in nested properties** – even when the parent struct's `physicalType` string does not directly contain `varchar`, soda-core compares against the real Databricks schema which uses `varchar(n)`, causing a mismatch

## Fix

Added `_has_unsupported_databricks_type()` which checks a column's `physicalType` and recurses into nested `properties` for any occurrence of `varchar` or `map<`. When a Databricks column matches, the schema type check is skipped and a `logger.warning()` is emitted instead of generating a failing check.

This is an interim fix until the Soda v3 dependency is removed entirely.

## Testing

- Manually verified against a real Databricks table with `varchar(30)`, `map<string,string>`, `struct<...:map<...>>`, and `struct<varchar_field:varchar(100)>` columns — all previously failing type checks now emit a warning and pass
- Unit tests added for `_has_unsupported_databricks_type` (11 parametrized cases) and for `to_schema_checks` behaviour on Databricks vs non-Databricks servers

Closes #1219
Closes #1245

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,7 +10,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Added
 
 ### Fixed
-
+- Schema type check no longer fails for `varchar(n)` columns on Databricks with PySpark 4.0+, and for `map` and `varchar` types nested inside `struct` columns; affected columns emit a warning and skip the type check instead
 
 ## [0.12.4] - 2026-05-21
 
--- a/datacontract/engines/data_contract_checks.py
+++ b/datacontract/engines/data_contract_checks.py
@@ -55,6 +55,26 @@ _ANSI_QUOTING_DIALECTS = {"postgres", "redshift", "sqlserver", "snowflake", "azu
 
 _BARE_IDENTIFIER_STRICT = re.compile(r"^[A-Za-z_][A-Za-z0-9_]*$")
 _BARE_IDENTIFIER_PERMISSIVE = re.compile(r"^[A-Za-z_][A-Za-z0-9_$]*$")
+
+_DATABRICKS_UNSUPPORTED_TYPE_RE = re.compile(r"\bvarchar\b|\bmap\s*<", re.IGNORECASE)
+
+
+def _has_unsupported_databricks_type(prop) -> bool:
+    """Return True if physicalType contains varchar(n) or map at any nesting level,
+    or if any nested property (recursively) has an unsupported type.
+
+    soda-core-spark-df ≤3.5.6 cannot match these types correctly on PySpark 4.0+:
+    - varchar(n): reported as a distinct VarcharType by PySpark 4.0+, not in SCHEMA_CHECK_TYPES_MAPPING
+    - map: convert_to_databricks() returns None for maps nested inside structs
+    Affected issues: #1245 (varchar), #1219 (map in struct)
+    """
+    if _DATABRICKS_UNSUPPORTED_TYPE_RE.search(prop.physicalType or ""):
+        return True
+    if prop.properties:
+        return any(_has_unsupported_databricks_type(nested) for nested in prop.properties)
+    return False
+
+
 _PERMISSIVE_BARE_DIALECTS = {"postgres", "redshift", "snowflake", "oracle", "sqlserver"}
 
 
@@ -138,9 +158,19 @@ def to_schema_checks(schema_object: SchemaObject, server: Server) -> List[Check]
 
         checks.append(check_property_is_present(schema_name, property_name, quoting_config))
         if check_types and (prop.physicalType is not None or prop.logicalType is not None):
-            sql_type: str = convert_to_sql_type(prop, server_type)
-            if sql_type is not None:
-                checks.append(check_property_type(schema_name, property_name, sql_type, quoting_config))
+            if server_type == "databricks" and _has_unsupported_databricks_type(prop):
+                logger.warning(
+                    "Skipping schema type check for column '%s': physicalType '%s' "
+                    "contains varchar(n) or map, which are not reliably matched by "
+                    "soda-core-spark-df ≤3.5.6 on PySpark 4.0+. "
+                    "See https://github.com/datacontract/datacontract-cli/issues/1245",
+                    property_name,
+                    prop.physicalType,
+                )
+            else:
+                sql_type: str = convert_to_sql_type(prop, server_type)
+                if sql_type is not None:
+                    checks.append(check_property_type(schema_name, property_name, sql_type, quoting_config))
         if prop.required:
             checks.append(check_property_required(schema_name, property_name, quoting_config))
         if prop.unique:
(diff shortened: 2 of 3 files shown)


IchEssBlumen:
Hi @jschoedl what do you think about this fix? We would skip physicalType varchars (like we do for maps) and emit a warning until we have a proper solution without the soda dependency. Would help us a lot, because we have many dataproduct using varchars and currently the contract validation fails for all of them if they are using pyspark 4.x and the latest datacontract-cli.
Closes https://github.com/datacontract/datacontract-cli/issues/1219
Closes https://github.com/datacontract/datacontract-cli/issues/1245

jschoedl:
LGTM! I don't think we need that many tests for this.

PR #1255, opened 2026-05-22
Feat/DatacontractException

### LINKED ISSUE #1251: feat/ for the schema and property DataContract Exception raise schema name and property name instead of index
Hi, 
Is it possible to improve the the DataContractException rendering

Image
replace the item position in error by its name properties .

turn `1) Check that data contract YAML is valid: data.schema[31].properties[86].description must be string` 
into `1) Check that data contract YAML is valid: data.schema.product.properties.category must be string`

dmaresma:
related to the issues #1251

- [x ] Tests pass (`uv run pytest`)
- [ x] Code formatted (`uv run ruff check --fix && uv run ruff format`)


jschoedl (changes requested):
Thank you!

Could you wrap the code in a try/catch block and fall back to the original error message if something goes wrong (maybe move the whole logic to a seperate function)? To not loose it completely, we could log errors from the conversion as `logger.info`.

PR #1261, opened 2026-05-26
feat: resolve authoritativeDefinitions[type=definition] and add --no-inline-references

andredlng:
## Summary

- Re-introduces the definition inlining that became a no-op in 0.11.x when the internal model moved from DCS to ODCS. For each schema property, `inline_definition_into_property` now resolves its first `authoritativeDefinitions[type=definition]` reference, parses the response as an ODCS `SchemaProperty`, and fills in the fields the property leaves unset. Inline values always win (`model_fields_set`); `authoritativeDefinitions`, `name`, `properties` and `items` are never overwritten.
- Relative `url:` is joined onto `ENTROPY_DATA_HOST`; absolute is used as-is. The `x-api-key` header is sent **only** when the resolved URL's host matches `ENTROPY_DATA_HOST` — a third-party `url:` cannot receive the user's key. Any failure (host mismatch, HTTP error, network error, malformed body) raises `DataContractException`; no silent skip. Per-process success-only cache so a shared definition is fetched once; failures aren't cached so a transient blip doesn't poison later runs.
- New `--no-inline-references / --inline-references` flag on `lint`, `test`, `ci`, `changelog`, and every `export` subcommand. Doubles as an escape hatch when a definition is briefly unreachable or invalid — the contract was already validated against the ODCS schema before resolution, so skipping the fetch still produces a syntactically valid run from the contract as written.
- Naming chosen generic (*references*, not *definitions*) so future `authoritativeDefinitions[type=…]` handlers can plug into the same flag and field. Existing `DataContract(inline_definitions=…)` parameter renamed to `inline_references`.
- Fixtures using `type: "definition"` against demo URLs switched to `type: "businessDefinition"` so the existing tests don't trigger HTTP; `shipments-odcs.xlsx` regenerated.
- 17 new tests in `tests/test_resolve_definitions.py` cover URL shapes, api-key scoping, inline-wins precedence, identity-field protection, nested properties / array items, caching, every failure mode, and a CLI smoke test proving `--no-inline-references` makes zero HTTP calls.


datacontract/lint/resolve.py
```diff
@@ -78,40 +80,144 @@ def resolve_data_contract(
...
     data_contract_str = read_resource(location)
-    return _resolve_data_contract_from_str(data_contract_str, schema_location, inline_definitions, all_errors)
+    return _resolve_data_contract_from_str(data_contract_str, schema_location, inline_references, all_errors)
+
+
+DEFINITION_AUTHORITATIVE_TYPE = "definition"
+
+# `name` is the property's own; `authoritativeDefinitions` is the link itself;
+# `properties`/`items` are the contract author's structure.
+_NON_MERGEABLE_FIELDS = frozenset({"name", "authoritativeDefinitions", "properties", "items"})
```
jschoedl, line 93:
```suggestion
# `name` and `id` are the property's own; `authoritativeDefinitions` is the link itself;
# `properties`/`items` are the contract author's structure.
_NON_MERGEABLE_FIELDS = frozenset({"id", "name", "authoritativeDefinitions", "properties", "items"})
```

datacontract/command_lint.py
```diff
@@ -36,12 +36,24 @@ def lint(
             help="Report all JSON Schema validation errors instead of stopping after the first one.",
         ),
     ] = False,
+    inline_references: Annotated[
+        bool,
+        typer.Option(
+            help="Resolve external references in the contract and inline the fetched content "
+            "from the configured entropy-data host (currently: authoritativeDefinitions[type=definition])."
```
jschoedl, line 43:
```suggestion
            help="Resolve external references (currently: authoritativeDefinitions\\[type=definition]) in the "
                 "contract and inline the fetched content from the configured entropy-data host."
```

Square brackets need to be escaped (otherwise they get interpreted as tags and do not show up in `datacontract lint --help`).

(update the other copies of this string as well)

CHANGELOG.md
```diff
@@ -8,6 +8,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## Unreleased
 
 ### Added
+- Resolve `authoritativeDefinitions[type=definition]` on schema properties: the referenced ODCS property is fetched from the configured entropy-data host (`ENTROPY_DATA_HOST`) and inlined into the property, filling fields the contract author left unset. The configured `x-api-key` is sent only when the resolved URL's host matches `ENTROPY_DATA_HOST` so a third-party `url:` cannot receive the user's key. Any resolution failure — host mismatch, HTTP error, network error, malformed body — rejects the contract.
+- `--no-inline-references` flag on `lint`, `test`, `ci`, `export`, and `changelog` to skip the HTTP fetch above and leave each property as written. Useful for offline runs (e.g. CI without `ENTROPY_DATA_API_KEY`) and pure round-trip exports (e.g. `export excel`).
```
jschoedl, line 12:
```suggestion
- Resolve `authoritativeDefinitions[type=definition]` on schema properties: the referenced ODCS property is fetched from the configured entropy-data host (`ENTROPY_DATA_HOST`) and inlined into the property, filling fields the contract author left unset. The configured `x-api-key` is sent only when the resolved URL's host matches `ENTROPY_DATA_HOST` so a third-party `url:` cannot receive the user's key.
- **Breaking:** Per default, any resolution failure of `authoritativeDefinitions[type=definition]` rejects the contract on `lint`, `test`, `ci`, `export`, and `changelog`.
- `--no-inline-references` flag to skip the HTTP fetch above and leave each property as written. Useful for offline runs (e.g. CI without `ENTROPY_DATA_API_KEY`) and pure round-trip exports (e.g. `export excel`).
```

jschoedl:
After updating help texts, please run `python3 update_help.py` to update the README.

andredlng:
Thanks for the review @jschoedl — all four points applied directly on main in 68249c2:

- `_NON_MERGEABLE_FIELDS` now includes `id` (with the comment updated)
- `\[type=definition]` escaped in the help text on `lint`, `test`, `ci`, `export`, and `changelog`; verified it renders in `datacontract lint --help`
- CHANGELOG restructured into three bullets with the breaking note on its own line
- `python3 update_help.py` re-ran; README regenerated

Sorry the PR was merged before your review came back — that was my mistake.

PR #1262, opened 2026-05-26
feat: resolve authoritativeDefinitions[type=semantics] (REST URL or IRI)

andredlng:
## Summary

Stacked on #1261. Broadens the resolver from `"definition"` only to a precedence-ordered list (`"semantics"`, `"semantic"`, `"definition"`), and routes IRI-shaped semantic URLs through a new entropy-data endpoint instead of trying to GET them directly.

### Routing rules

| `url:` host | `type` | What happens |
|---|---|---|
| Matches configured entropy-data host (relative or absolute) | any | Fetched directly; `x-api-key` sent when configured |
| Off-host | `definition` | Fetched directly and anonymously — a contract may legitimately reference a third-party REST URL, and the API key must never leak across hosts (existing behaviour, unchanged) |
| Off-host | `semantics` / `semantic` | Treated as an IRI; routed through `GET /api/semantics?iri=...` on the configured entropy-data host (new endpoint in the entropy-data PR). Requires an API key — that endpoint is API-key only |

### Precedence

A property that links to both a semantics and a definition reference resolves through the **semantic** one — matches `useInheritedDefinition` in `datacontract-editor`. Legacy singular `"semantic"` is also accepted for contracts written before the entropy-data type migration.

### Tests

Five new tests in `test_resolve_definitions.py` (17 → 22): REST URL with api-key, IRI routed through `/api/semantics`, IRI without api-key raises with clear guidance, legacy `"semantic"` type resolves, semantics-wins-over-definition.

Verified end-to-end against a local entropy-data instance: contract with one REST-URL semantic and one IRI semantic — both resolve and inline correctly, no spurious requests to the IRI's host.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,7 +9,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 - Resolve `authoritativeDefinitions[type=definition]` on schema properties: the referenced ODCS property is fetched from the configured entropy-data host (`ENTROPY_DATA_HOST`) and inlined into the property, filling fields the contract author left unset. The configured `x-api-key` is sent only when the resolved URL's host matches `ENTROPY_DATA_HOST` so a third-party `url:` cannot receive the user's key.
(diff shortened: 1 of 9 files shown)


jschoedl (approved):
LGTM!

PR #1281, opened 2026-06-04
Bundle Spark Kafka/Avro connector JARs for air-gapped Kafka tests

jochenchrist:
## Problem

The first `datacontract test` against a Kafka server downloads the Spark Kafka + Avro connector JARs (~60 MB) from Maven Central at runtime, because `create_spark_session()` always sets `spark.jars.packages`. That needs outbound network and adds roughly a minute to startup, so the image is not air-gapped.

A warm Ivy cache does **not** fix this. Spark hands the coordinates to Ivy on every session start, and Ivy still contacts a resolver. With no network it fails:

```
:: UNRESOLVED DEPENDENCIES ::
:: org.apache.spark#spark-sql-kafka-0-10_2.13;4.0.2: not found
:: org.apache.spark#spark-avro_2.13;4.0.2: not found
```

## Fix

1. **Dockerfile**: pre-fetch the connectors at build time and bundle them onto the PySpark classpath. They are resolved with the exact Spark/Scala version of the bundled PySpark (`spark_connector_packages()`), so they always match the runtime, and renamed to their canonical Maven filenames.
2. **`kafka.py`**: `create_spark_session()` skips `spark.jars.packages` when the connectors are already on the classpath (`connector_jars_already_on_classpath()`). Pip-installed users without the bundled JARs keep the download-on-first-use behaviour, so nothing changes outside the image.

## Verification

Built a test image applying both changes on top of the published image (the real Dockerfile uses DHI base images that need `dhi.io` access).

- **Build** resolves and bundles 13 connector JARs onto the classpath.
- **Offline session**: `create_spark_session()` starts with `--network none` (no Maven route): `BUNDLED_JARS_DETECTED: True`, `SESSION_OK 4.0.2`.
- **End-to-end on an internal network with zero internet egress** (Maven unreachable), against a real KRaft Kafka broker: `🟢 data contract is valid. Run 16 checks. Took 5.4 seconds` (vs ~70s with the Maven download on the current image).

## Notes

- The prefetch step runs in the runtime stage because it needs the JRE that is copied there; it adds ~60 MB to the image (the connector JARs that were downloaded at runtime before).
- Image size grows by the connector JAR set; in exchange the Kafka engine starts offline and ~13x faster.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## Unreleased
 
+### Changed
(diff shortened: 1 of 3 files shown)


jochenchrist:
too complicate Dockerfile

PR #1285, opened 2026-06-08
Make agent instructions vendor-agnostic via AGENTS.md

wardgielis:
Move project instructions from `CLAUDE.md` to `AGENTS.md` so they work with any AI coding agent (opencode, Claude Code, Cursor, etc.). `CLAUDE.md` now points to `AGENTS.md` as the canonical source.

This follows the emerging best practice where `AGENTS.md` is the single source of truth for agent instructions, with tool-specific files (like `CLAUDE.md`) referencing it.


jschoedl:
We should use `@AGENTS.md` to import the agents file. Otherwise, Claude does not necessarily read it.

PR #1292, opened 2026-06-09
feat(bigquery): support separate billing project for cross-project validation

### LINKED ISSUE #1291: BigQuery: no way to specify a separate billing/job project — validation always submits jobs to the data project
When running datacontract test against a BigQuery server where the data lives in an external project (e.g. a source org's project), the CLI submits the Soda scan jobs directly to that same project. There is no way to configure a separate billing/compute project for job execution.

This makes cross-project and cross-organisation validation setups impossible without granting bigquery.jobUser on the source project — which is undesirable from a billing isolation and least-privilege standpoint.

**Error**

403 POST https://bigquery.googleapis.com/bigquery/v2/projects/{source-project}/queries?prettyPrint=false:
Access Denied: Project <source-project>: User does not have `bigquery.jobs.create` permission in project {source-project}.

**Expected behaviour**

BigQuery natively supports cross-project query execution: a job runs and bills in Project A while reading data from Project B, as long as the identity has bigquery.jobUser on A and bigquery.dataViewer on B. The CLI should support this pattern.

A proposed addition to the server config:

```yaml
servers:
  production:
    type: bigquery
    project: source-org-project      # where the data lives
    dataset: core_dataset
    billing_project: my-own-project  # where jobs run and are billed
```

If billing_project is set, the BQ client should be initialised with project=billing_project so that query jobs are submitted there, while the FROM clause references the data project via fully qualified names (source-org-project.core_dataset.table).

Workaround
Currently none — granting `bigquery.jobUser` on the source project is the only option, which bills the source org and violates least-privilege.

Environment
datacontract-cli version: 0.10.x
Engine: soda-core-bigquery
Auth: Application Default Credentials (Cloud Composer SA)

sibajikani:
Closes #1291

## Problem

When running `datacontract test` against a BigQuery server in an external project,
the CLI submits Soda/ibis query jobs directly to that project — requiring
`bigquery.jobUser` on the data source project. This makes cross-project and
cross-organisation validation impossible without granting the running identity
compute permissions on a project it doesn't own, which also routes billing charges
to the wrong project.

## Solution

Added support for a separate billing/compute project via:

1. **Environment variable:** `DATACONTRACT_BIGQUERY_BILLING_PROJECT_ID`
2. **Per-contract custom property:**
   ```yaml
   servers:
     production:
       type: bigquery
       project: source-org-project    # data lives here
       dataset: core_dataset
       customProperties:
         - property: billingProject
           value: my-ingestion-project  # jobs run and bill here
When a billing project is configured and differs from the data project, a
google.cloud.bigquery.Client is initialised with project=billing_project
and passed to ibis.bigquery.connect(). This means:

Jobs are submitted to (and billed against) the billing project → requires bigquery.jobUser there
Table data is read from the data project (server.project) → requires bigquery.dataViewer there
The custom property takes precedence over the env var. If neither is set,
or if the billing project equals the data project, behaviour is unchanged.

## Changes

datacontract/engines/ibis/connections/connect.py — billing project routing in the bigquery connection branch
README.md — new env var documented in the BigQuery environment variables table
CHANGELOG.md — entry under Unreleased

## Testing

Existing BigQuery tests pass (integration tests are skipped without
DATACONTRACT_BIGQUERY_ACCOUNT_INFO_JSON_PATH as before). The billing project
path was validated manually against a cross-project BigQuery setup. Full suite:
758 passed, pre-existing failures unrelated to this change (Docker/Java/Spark
not available in local environment).
Add support for billing project in BigQuery connection

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,7 +8,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## Unreleased
 
 ### Added
-- BigQuery: added support for a separate billing/compute project via the `DATACONTRACT_BIGQUERY_BILLING_PROJECT_ID` environment variable or a `billingProject` server custom property. When set, query jobs are submitted to (and billed against) the billing project while data is read from the `project` specified in the server config. This enables cross-project and cross-organisation validation without requiring `bigquery.jobUser` on the data project.
+- BigQuery: added support for a separate billing/compute project via the `DATACONTRACT_BIGQUERY_BILLING_PROJECT` environment variable. When set, query jobs are submitted to (and billed against) the billing project while data is read from the `project` specified in the server config. This enables cross-project and cross-organisation validation without requiring `bigquery.jobUser` on the data project.
 
 ## [1.0.0] - 2026-06-04
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## Unreleased
 
+### Added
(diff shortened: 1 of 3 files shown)


sibajikani:
@jochenchrist made the update. Only using the env var now.

Issue #368, opened 2024-08-08
Add export for ClickHouse with --format clickhouse


lideke:
Hello,

Expand the support for ClickHouse by adding  export functionality 

Thanks

jochenchrist:
Also `test` would be nice for full clickhouse support.

jochenchrist:
Required Data Contract Specification Updates to support clickhouse is scheduled for V1.2.0:
https://github.com/datacontract/datacontract-specification/pull/111

jochenchrist:
ODCS 3.1.0 has support for clickhouse, this feature could now be implemented.

wardgielis:
Perfect! I will give it a try and create a Pull request

wardgielis:
@jochenchrist do you think it would be best to build it as a sql export with a sql-server-type option as clickhouse, or would a dedicated export format clickhouse make more sense?

jochenchrist:
Not a ClickHouse expert, but this would look fine for me:

```
datacontract export sql --dialect clickhouse --output output.sql datacontract.yaml 
```

wdyt?

wardgielis:
Looks good to me as well, I will see if I can implement it like this

wardgielis:
Implemented the export in https://github.com/datacontract/datacontract-cli/pull/1293 

Will create a seperate PR for the import

jochenchrist:
implemented with https://github.com/datacontract/datacontract-cli/pull/1293

PR #1293, opened 2026-06-09
Add ClickHouse dialect support to SQL export

wardgielis:
Add ClickHouse as a supported SQL dialect for `datacontract export sql`.

### Changes

- **Type mapping**: `convert_type_to_clickhouse()` with full ClickHouse type support (`String`, `UUID`, `DateTime64(6)`, `Decimal(P,S)`, `Float32/64`, `Int8/16/32/64`, `Bool`, `Map(K,V)`, `Tuple(...)`, `Array(T)`, `FixedString(N)`, plus custom `clickhouseType` override)
- **Nullable handling**: Required fields use bare types (`String`), optional fields use `Nullable(String)` — idiomatic ClickHouse
- **Identifiers**: Backtick-quoted column and table names
- **DDL generation**: `CREATE TABLE IF NOT EXISTS`, column/table comments
- **Sorting key**: ClickHouse uses `ORDER BY` as its sorting/primary key mechanism. Primary key fields in the data contract are automatically translated to the `ORDER BY` clause (no inline `PRIMARY KEY`). Overridable via `--clickhouse-order-by`.
- **Engine**: Defaults to `ENGINE = MergeTree()`, overridable via `--clickhouse-engine`
- **Auto-detect**: Automatically detects ClickHouse dialect from `type: clickhouse` in the data contract servers block
- **CLI**: `--dialect clickhouse`, `--clickhouse-engine`, `--clickhouse-order-by` options
- **Bug fixes**: Server loop in `to_sql_ddl` no longer overrides explicit `--dialect` with first server type found; table prefix only applied when target dialect matches server type. Composite PK constraint (`CONSTRAINT ... PRIMARY KEY`) is skipped for ClickHouse (invalid syntax).
- **Tests**: 5 ClickHouse-specific tests + all existing SQL export tests pass unchanged

### Documentation

- README updated with ClickHouse type mapping, usage examples, engine/ORDER BY docs (noting ORDER BY serves as the primary key), and updated DDL example
- CHANGELOG entry under Unreleased


wardgielis:
Implements https://github.com/datacontract/datacontract-cli/issues/368 

will do the import in a seperate Pull Request to not make it too big

jschoedl (approved):
Thank you for the PR! I directly applied a few minor changes, hope that's okay :)

wardgielis:
> Thank you for the PR! I directly applied a few minor changes, hope that's okay :)

Nice, I had a look at your changes and they are great improvements indeed, thanks! :D

Issue #1280, opened 2026-06-04
Inconsistency in datacontract types between import and test


PaulMag:
Hello team!

We are using `datacontract import unity` to generate datacontracts for our tables in Databricks. Later we want to validate the tables against these contracts using `datacontract test`.

It mostly works as expected, except for columns with complex types: struct and array. For these types the generated contract is not compatible with what the validation expects. 

EDIT: These tests were made on `v0.12.2`. Below is "Update after v1.0.0" (does not solve the issue).

# Reproduce problem

In a Databricks notebook:

```SQL
CREATE TABLE inconsistency_example (
  id BIGINT,
  id_list ARRAY<BIGINT>,
  id_struct STRUCT<value: BIGINT>
);
```

Import a contract for that table with datacontract-cli:

```
export DATACONTRACT_DATABRICKS_TOKEN=<your token>
export DATACONTRACT_DATABRICKS_SERVER_HOSTNAME=<your hostname>
export DATACONTRACT_DATABRICKS_HTTP_PATH=<your http path>

datacontract import unity --table your_catalog.your_schema.inconsistency_example --output /tmp/generated_contract.yaml
```

```yaml
#generated_contract.yaml
(...)
schema:
- name: inconsistency_example
  physicalType: table
  businessName: inconsistency_example
  logicalType: object
  physicalName: inconsistency_example
  properties:
  - name: id
    physicalType: bigint
    logicalType: integer
  - name: id_list
    physicalType: array<bigint>
    logicalType: object
  - name: id_struct
    physicalType: struct<value:bigint>
    logicalType: object
```

Validate the generated contract with datacontract-cli:

```
$ datacontract test /tmp/generated_contract.yaml
(...)
🔴 data contract is invalid, found the following errors:
1) id_list Check that field id_list has type ARRAY<STRING>: Type Mismatch, Expected Type: ARRAY<STRING>; Actual Type: array<bigint>
2) id_struct Check that field id_struct has type STRUCT<>: Type Mismatch, Expected Type: STRUCT<>; Actual Type: struct<value:bigint>
```
It fails. 😔

# Workaround

To get the validation to pass, we have to manually rewrite the types.
Instead of the entire composite type in a string like `physicalType: struct<value:bigint>`, the datacontract needs the type to be structured yaml. 
`struct<...>` must be replaced with a `properties:` element, and `array<...>` must be replaced with an `items:` element.
[...]

PR #1300, opened 2026-06-12
fix: import unity struct and array columns as structured ODCS types

jochenchrist:
Fixes the import side of #1280.

## Problem

`datacontract import unity` emitted complex column types only as a flat Spark DDL string (`physicalType: struct<value:bigint>`, `array<bigint>`) with `logicalType: object`, while `datacontract test` and the ODCS spec expect the structured representation (nested `properties` for structs, `items` for arrays, `logicalType: array` for arrays).

## Changes

- The unity importer now parses Unity's `type_json` (Spark StructField JSON; `StructField.fromJson` is pure Python, no JVM or SparkSession needed) and emits nested `properties` for struct columns and `items` for array columns, reusing the spark importer's recursive conversion. Nested field comments are carried over. If pyspark is unavailable or `type_json` is missing/unparseable, it logs a warning and falls back to the previous flat output instead of failing the import.
- `map_type_from_sql` now maps `array<...>` to `logicalType: array` (previously `object`), with explicit `struct`/`map` branches. This also corrects the `sql` and `snowflake` importers, which share the function.
- Map columns keep the flat `map<k,v>` string in `physicalType` with `logicalType: object` for now: ODCS v3.1 has no map representation. Once ODCS v3.2 lands (`logicalType: map` with `map.key`/`map.value`, [RFC 0030](https://github.com/bitol-io/tsc/blob/main/rfcs/approved/odcs-v3.2.0/0030-maps.md)) and the model library supports it, the importer can emit structured maps from the already-parsed `MapType` in the same helper.

## Tests

New fixture modeled on the issue's repro table (`bigint`, `array<bigint>`, `struct<value:bigint>`, plus `array<struct<...>>` and `map<string,bigint>`) with a test asserting the structured output and that the result passes lint. The issue's example table now imports as exactly the contract the reporter had to write by hand.

Note: this does not change the test engine's type checking, which still compares only top-level type categories (the v1.0.x part of the issue). The imported contracts now carry the nested type information, so deeper validation can build on this.


datacontract/imports/sql_importer.py
```diff
@@ -358,6 +358,13 @@ def map_type_from_sql(sql_type: str) -> tuple[str, str | None]:
         return ("string", None)
     elif sql_type_normed == "clob" or sql_type_normed == "nclob":
         return ("string", None)
+    elif sql_type_normed.startswith("array"):
+        return ("array", None)
+    elif sql_type_normed.startswith("struct"):
+        return ("object", None)
+    elif sql_type_normed.startswith("map"):
+        # ODCS v3.1 has no map logical type; RFC 0030 adds logicalType: map in v3.2
+        return ("object", None)
```
PaulMag: Is it better to just return `None` instead of `"object`" here until we get the `"map"` option? `"object"` is not a valid `logicalType` for maps. When running `datacontract test` with a map field I get
```
Expected type 'object' (struct) but column is 'map<..., ...>' (map)
```
So for map fields I have to just completely remove `logicalType` to pass validation.
This doesn't matter if `physicalType` is also present, but it seems fragile to rely on an invalid parameter being ignored due to the presence of a different parameter.

The same arguemnt might go for the `else:` clause as well. As I understand only `struct` types successfully validates as `"object"`.

jochenchrist: Yes good comment.
@copilot please change

jochenchrist:
@copilot resolve the merge conflicts in this pull request

datacontract/imports/sql_importer.py
jochenchrist: @copilot  please changet just return None instead of "object" here accordingly

PR #1301, opened 2026-06-12
feat(trino): support jwt and oauth2 authentication for Trino

### LINKED ISSUE #1138: Support OAuth2 / JWT auth for Trino connections
# Description

The current Trino connection mapping in datacontract-cli seems to only expose basic auth-related fields such as username/password, but not Trino auth modes like OAuth2 or JWT. 

Source: https://github.com/datacontract/datacontract-cli/blob/main/datacontract/engines/soda/connections/trino.py

This makes the Trino integration hard to use in environments where password auth is not available and OAuth2 or JWT is required.

# Suggested improvement

Extend the Trino connection handling so auth settings can be passed through to the underlying Soda layer, especially for:

- OAuth2
- JWT

I’d be happy to contribute a PR.

hk8suva:
Add JWT and OAuth2 authentication support for Trino via DATACONTRACT_TRINO_AUTHENTICATION environment variable.

Fixes #1138

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 - `datacontract import powerbi` imports a data contract from a Power BI semantic model (.pbit, .bim, or .json) (#1233)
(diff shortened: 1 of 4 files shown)


jschoedl (approved):
Thank you! LGTM.

PR #1311, opened 2026-06-16
1278 recursive databricks and spark checks  and custom queries

rob-h-w:
Supports recursive databricks & spark checks and custom queries, replaces the ibis implementation that requires write permissions on connection until a new release of ibis stops doing that.

- [-] README.md updated (if relevant)

I don't see a `README.md` update that's in scope.


jochenchrist:
@rob-h-w I see this PR is still in draft. Give me a ping whenever you feel this is ready to review.

rob-h-w:
> @rob-h-w I see this PR is still in draft. Give me a ping whenever you feel this is ready to review.

Hi, @jochenchrist, I think this is worth a look now. Apologies for the delay.

jochenchrist:
The root cause is fixed with https://github.com/ibis-project/ibis/pull/11962 (yet not released), correct?

datacontract/engines/ibis/connections/connect.py
```diff
@@ -105,8 +108,26 @@ def connect_ibis(
             database_name = ".".join(filter(None, [server.catalog, server.schema_]))
             if database_name:
                 spark.sql(f"USE {database_name}")
+            from datacontract.engines.ibis.connections.kafka import add_spark_nested_views_for_contract
```
jochenchrist, line 111: We should not have a kafka package dependency in the non-kafka path.

mahic:
@jochenchrist Ibis has an upcoming release https://github.com/ibis-project/ibis/issues/12102 which probably would include the https://github.com/ibis-project/ibis/pull/11962 fix

README.md
```diff
@@ -349,6 +349,21 @@ Try to sync all groups:
...
+sudo apt-get install -y libpq-dev
+```
+**Fedora/RHEL:**
+```bash
+# Fedora/RHEL:
+sudo dnf install -y postgresql-devel
+# Arch:
+sudo pacman -S postgresql-libs
+```
+
```
jschoedl, line 366: This should no longer be needed on current main, afaik.

jschoedl:
Hi @rob-h-w, thank you for your PR!

As far as I can tell, this PR is not really blocked by the Ibis release since the workaround for the Ibis bug is already at `main` anyways and doesn't get introduced here.

I took a look at the code and took the liberty to change a few parts of the approach, although the main part stayed. Most importantly, you had added `myModel__items` models which were then made queryable using a Spark temp view or SQL strings attached to a Databricks connection. Now, `model` names only the real table, and we instead compile a predicate like `SIZE(FILTER(items, i -> i.sku IS NULL)) > 0` over the elements.

This seems a bit more robust to me than messing with the Databricks connection. It also fixed a few things I ran into while testing:

 - an order with an empty items array failed `required: true` on an item field, because explode_outer produces a NULL row for it
- `--filter` was silently ignored for the nested checks, since orders__items is never a key in the filter map
- counts came back in item rows while every other check counts table rows, and both get summed into one quality score

PR #1331, opened 2026-06-22
feat: Recursive Validation of `logicalType`

### LINKED ISSUE #1280: Inconsistency in datacontract types between import and test

sondrfos:

fixes: #1280

See context here: https://github.com/datacontract/datacontract-cli/issues/1280#issuecomment-4688175591
We have followed up and solved issue number 3 in the list:

> 3. Since v1.0.0 the test engine only compares top-level type categories, so element types of arrays, structs, and maps are not validated. Your observation is correct. This was a consequence of the soda-to-ibis migration. With the importer now emitting structured nested types, recursive type checking has something to validate against; I'll track that as a separate enhancement.

We have mapped ibis types to `SchemaProperty` objects in order to perform a deep type check on objects and arrays. 

> physicalType is an opaque platform-specific string, useful for documentation and platform-specific tooling, and is not the basis for nested validation.

This PR also addresses this point, please let us know if this should be addressed in a separate PR. 

> For maps: ODCS v3.1 has no structured map representation, which is why the flat map<k,v> string was the only option. ODCS v3.2 adds logicalType: map with map.key/map.value ([RFC 0030](https://github.com/bitol-io/tsc/blob/main/rfcs/approved/odcs-v3.2.0/0030-maps.md)). Once that lands in the model library, the importer will emit structured maps too; until then maps keep the flat physicalType string.

Finally we have mapped unsupported types to None for the logicalType field. This is a breaking change if anyone depends on this functionality, but we understood your comment that this should not be supported until v3.2.0 

## Further Clarifications

For objects and arrays we have changed the error message which is presented to the user to specify the first failed nested field type check. There are several ways to approach this: 

1. Show the first failed field in the nested structure
2. Show all failed fields in the nested structure
3. Show the first failed field and a count of the number of additional errors
4. Show a flat string representation of the object as shown in  #1280 

We would recommend approach number 3, since the other approaches become quite impractical for large nested structures. See the example below

```
╭────────┬────────────────────────────────────────────┬───────────┬────────────────────────────────────────────────────╮
│ Result │ Check                                      │ Field     │ Details                                            │
├────────┼────────────────────────────────────────────┼───────────┼────────────────────────────────────────────────────┤
│ failed │ Check that field example has type array    │ example   │ field '[].metadata.item[]': expected type          │
│        │                                            │           │ 'integer' but got 'string' (and 1 other type error)│
│ passed │ Check that field 'example' is present      │ example   │                                                    │
╰────────┴────────────────────────────────────────────┴───────────┴────────────────────────────────────────────────────╯
```

Here is an example of a very complex structure. We have added an error in the `example.metadata.item` and the `example.updated.timestamp` field as shown above. 

```
schema:
- name: abc
  physicalType: table
  physicalName: inconsistency_example
  properties:
  - name: example
    logicalType: array
    items:
      logicalType: object
      properties:
      - name: codeId
        logicalType: object
        properties:
        - name: value
          physicalType: bigint
          logicalType: integer
      - name: id
        physicalType: bigint
        logicalType: integer
      - name: metadata
        logicalType: object
        properties:
        - name: item
          logicalType: array
          items:
            physicalType: string
            logicalType: integer
      - name: updated
        logicalType: object
        properties:
        - name: timestamp
          physicalType: string
          logicalType: integer
      - name: updatedBy
        physicalType: string
        logicalType: string
      - name: originCodeId
        logicalType: object
        properties:
        - name: value
          physicalType: bigint
          logicalType: integer
      - name: uuid
        logicalType: object
        properties:
        - name: namespace
          physicalType: string
          logicalType: string
        - name: uuid
          physicalType: string
          logicalType: string
      - name: versionId
        physicalType: bigint
        logicalType: integer
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,13 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Fixed
+- `datacontract test` now only supports logicalTypes. Previously physicalType was preferrerd and used even if logicalType did not exist. 
+- `datacontract test` field type check now compares the full structured type tree for `object` and `array` logical types.
+- Unknown and unsupported types are silently ignored rather than failing the check. Specifically the `map` type is not supported until ODCS version v3.2.0 and is also ignored. 
+
+
+
 ## [1.0.4] - 2026-06-22
 
 ### Added
--- a/datacontract/engines/checks/check_spec.py
+++ b/datacontract/engines/checks/check_spec.py
@@ -14,7 +14,10 @@ from __future__ import annotations
 
 from dataclasses import dataclass
 from enum import Enum
-from typing import Any, List, Optional
+from typing import TYPE_CHECKING, Any, List, Optional
+
+if TYPE_CHECKING:
+    from open_data_contract_standard.model import SchemaProperty
 
 
 class MetricType(str, Enum):
@@ -116,8 +119,9 @@ class CheckSpec:
     valid_min_length: Optional[int] = None  # INVALID_COUNT
     valid_max_length: Optional[int] = None  # INVALID_COUNT
 
-    expected_category: Optional[str] = None  # FIELD_TYPE: normalized type category
+    expected_category: Optional[str] = None  # FIELD_TYPE: human-readable label (display only)
     expected_type_label: Optional[str] = None  # FIELD_TYPE: human-readable expected type
+    expected_schema_property: Optional["SchemaProperty"] = None  # FIELD_TYPE: structural comparison
 
     columns: Optional[List[str]] = None  # DUPLICATE_COUNT across multiple columns
 
--- a/datacontract/engines/checks/create_checks.py
+++ b/datacontract/engines/checks/create_checks.py
@@ -53,12 +53,6 @@ def to_schema_name(schema_object: SchemaObject, server_type: Optional[str]) -> s
     return schema_object.name
 
 
-def expected_type_category(prop: SchemaProperty) -> tuple[SchemaProperty, str]:
-    """Return (schema property for structural type comparison, human label)."""
-    label = prop.physicalType or prop.logicalType
-    return prop, (label or "")
-
-
 _PERCENT_UNITS = {"percent", "percentage", "%"}
 
 
@@ -177,8 +171,8 @@ def _to_schema_checks(schema_object: SchemaObject, server: Optional[Server]) -> 
             )
         )
 
-        if check_types and (prop.physicalType is not None or prop.logicalType is not None):
-            schema_prop, label = expected_type_category(prop)
+        if check_types and prop.logicalType is not None:
+            schema_prop, label = prop, prop.logicalType or ""
             checks.append(
                 CheckSpec(
                     key=f"{model}__{field}__field_type",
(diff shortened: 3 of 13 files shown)


jochenchrist:
LGTM

jochenchrist:
Can you also update the changelog?

PR #1332, opened 2026-06-22
Feat/hana engine

ToniLippmann:
Implements [RFC-0045: SAP HANA server type](https://github.com/bitol-io/tsc/blob/main/rfcs/0045-hana-server-type.md).

Adds a new `hana` server type for testing data contracts against
**SAP HANA Cloud** and **SAP Datasphere** (Open SQL schemas).

- Native engine via `hdbcli` (SAP HANA Client) — bypasses the Ibis/sqlglot
  path because neither supports SAP HANA
- Follows existing server-type conventions (`host`, `port`, `schema`;
  credentials via `DATACONTRACT_HANA_*` env vars)
- Datasphere: validates `primaryKey`/`unique` at the **data level**
  (`COUNT(DISTINCT ...)`) because Open SQL objects are predominantly views
  with no catalog constraint metadata
- `[hana]` is an **opt-in extra only** — `hdbcli` is distributed under SAP's
  proprietary license (SAP Developer License) and is therefore intentionally
  excluded from `[all]`

## Test results

Unit tests (mocked `hdbcli`, no live instance required):
- `test_hana_connection.py` — passed
- `test_hana_schema_check.py` — passed
- `test_hana_quality_check.py` — passed
- `test_hana_type_mapping.py` — passed
- `test_test_hana.py::test_hana_test_flow_bypasses_soda` — passed

Integration tests (SAP Datasphere demo tenant):
- `test_hana_full_contract_pass` — passed
- `test_hana_full_contract_fail_schema` — passed
- `test_hana_quality_check_live` — passed


simonharrer:
Only works with ODCS 3.2, as we are adding hana as a server type

jochenchrist:
@simonharrer Do we want to risk to depend on an non OSS license?

simonharrer:
@jochenchrist yes, but excluded via extra all. So none will install it except they really need that. We have the same issue with Teradata, and probably other vendors as well.

jochenchrist:
ok, then let's go with SAP HANA :)

jochenchrist (approved):
Approved. Pushed a follow-up that makes the HANA engine honor --filter/--filters, --dimension, --quality-id and --tag (they were silently ignored), carries the quality rule metadata on its checks, keys the service level checks per field, declares the DATACONTRACT_HANA_* variables as Config fields, and adds the SAP HANA connection guide.

jochenchrist:
Thanks for your improvement :)

PR #1334, opened 2026-06-23
Error message with recursive error count

jorgengranseth:
- [ ] ~~CHANGELOG.md entry added~~ we can add a changelog entry if this is considered sufficiently important

This is a completion of the suggestion outlined in #1331 under "Further Clarifications":
> Show the first failed field and a count of the number of additional errors

as well as a test case to illustrate and verify correct behaviour.

A curious observation when writing this test is that Ibis uses the physical types from the data contract as the source of truth for the parquet schema, rather than parsing the parquet file. Importing a parquet file also yields only top-level typing for nested types (object and array). It seems like `pyarrow.parquet` is able to parse types correctly:
```python
import pyarrow.parquet as pq

schema = pq.read_schema("tests/fixtures/parquet/data/nested_type_structures.parquet")
print(schema)
```
```
id: int64
values: list<element: int64>
  child 0, element: int64
metadata: struct<active: bool, name: string>
  child 0, active: bool
  child 1, name: string
```
So there may be room for improvement on the import side as well.


jschoedl:
Thanks, LGTM!

Issue #1250, opened 2026-05-21
Support programmatic configuration variables


mahic:
The library supports setting configuration variables such as `DATACONTRACT_DATABRICKS_TOKEN` via `os.environ`
In a multithreaded environment, this will cause race condition issues and security issues if the token supplied is a user token.

The workaround currently is to invoke subprocesses with env vars set for each sub process, but it would be a far better solution if one could set configuration vars programmatically, ie as a param on the `DataContract` type.

jschoedl:
Sounds good!

We plan to change our checks library soon (from Soda to Ibis). This might be a good opportunity to add this, because we'll touch each testing backend anyways.

jschoedl:
With version 1.0, we dropped Soda, which should make this easier. Would appreciate a PR :)

jschoedl:
Btw, with https://github.com/datacontract/datacontract-cli/pull/1295, the configuration variables can now also be provided using a .env file, which might be a bit more convenient than the current behavior in some cases.

jochenchrist:
I think this is a very valid request.
We should support setting configuration variables via `kwargs` or a dedicated configuration object as an alternative way to set credentials and other vars. Also for `datacontract api` there is the need to specify credentials via HTTP headers.

jorgengranseth:
We took a stab at implementing programmatic configuration with env vars as backup in #1339. Is that the direction you were thinking, @jochenchrist?

jschoedl:
Hi @jorgengranseth, thank you again for your PR! And sorry for the delay.

After discussing this with Jochen, we decided that it might be better to use Pydantic Settings for this. I'm currently working on preparing a draft to figure how that would look like in comparison.

jochenchrist:
Resolved with v1.0.17
https://docs.datacontract.com/configuration#python-library

PR #1339, opened 2026-06-24
Pass Databricks config programmatically

jorgengranseth:

An approach to #1250 with Databricks as a proof of concept.

`UnityImporter` did not support Application Default Credentials, which the Databricks `WorkspaceClient` picks up by default, failing the command if both ADC and another method (required by the CLI) were provided. I changed the code to allow ADC to be used, but it would be better if it were possible to override ADC on-demand with this custom configuration. It is a bit unclear to me whether that is an issue for this repo or databricks-sdk. Ibis seems to handle this fine, using `databricks.sql` rather than the `WorkspaceClient`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - `datacontract test` field type check now compares the full structured type tree for `object` and `array` logical types.
 - Unknown and unsupported types are silently ignored rather than failing the check. Specifically the `map` type is not supported until ODCS version v3.2.0 and is also ignored. 
 
+### Added
+- The `test` and `import` commands support programmatically overriding Databricks configuration, otherwise passed as env variables.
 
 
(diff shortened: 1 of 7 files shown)


jochenchrist:
Thanks for the PR.
I need a few days to evaluate this approach vs alternatives, and if we can/want to extend this configuration to all other data sources.

jochenchrist:
Thank you for this PR and for pushing the topic forward — it directly shaped the discussion in #1250. Closing it now, since the programmatic configuration part has shipped in a generalized form: #1467 introduced a typed `Config` class (Pydantic Settings) covering all data sources, usable via `DataContract(config=...)`, a YAML config file, or per-request API headers, with env variables as fallback. See the new documentation: https://docs.datacontract.com/configuration

The second part of your PR, falling back to the Databricks SDK's Application Default Credentials in the Unity importer, is still a good idea and not yet on main. A fresh PR implementing that cascade on top of the new `Config` system would be very welcome.

PR #1357, opened 2026-07-01
feat: include datacontractCliVersion in test run JSON

### LINKED ISSUE #1353: Add CLI version to test result JSON
### Description
The `datacontract test` command can write test results as JSON using `--output` / `--output-format json`.

We use this JSON output as machine-readable input to track metrics over multiple test runs. For that use case, we parse and persist fields from the JSON output.

Currently, the JSON output does not seem to include:

- the `datacontract-cli` version that produced the file
- a dedicated JSON output schema/version field

This makes it hard to safely handle future changes to the JSON structure. If the output format changes between CLI versions, downstream parsers need to know which structure they are parsing.

### Possible Solution

Adding the `datacontractCliVersion` field to the test result JSON.

I would be happy to contribute a PR if this change makes sense for the project.

hk8suva:
Fixes #1353 

This PR only adds the datacontract cli version to the test output.
It does not add a dedicated JSON output schema/version field. If that would be useful as well, I can try to add it.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Added
+- `datacontract test` JSON output now includes `datacontractCliVersion` (#1353)
+
 ### Fixed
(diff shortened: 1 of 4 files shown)


jschoedl (approved):
Thanks, LGTM!

PR #1372, opened 2026-07-09
feat(1367): add support for BigQuery INTERVAL type on import

### LINKED ISSUE #1367: Import fails for BigQuery INTERVAL type
When running the `import` command for BigQuery (`datacontract-cli` 1.x), the command fails if the table contains column of the `INTERVAL` type.

fantastisch:
Adds the `INTERVAL` type to the list of supported physical types for BigQuery.
~Also, removes `INT64` and `FLOAT64`, as these values can't be imported from BigQuery (the returned values for these types are `INTEGER` and `FLOAT`, which are already in the mapping list.~


datacontract/imports/bigquery_importer.py
```diff
@@ -274,6 +274,7 @@ def map_type_from_bigquery(bigquery_type_str: str) -> str:
         "BIGNUMERIC": "number",
         "GEOGRAPHY": "object",
         "JSON": "object",
+        "INTERVAL": "number",
```
jschoedl: Is there a reason for logicalType `number` for intervals? As they are represented in a form like `1-2 3 4:5:6.789`, I'd guessed that `string` fits better.

jschoedl:
Follow-up issue for this change would be to make `test` work with INTERVALS, which is currently broken. :)

datacontract/imports/bigquery_importer.py
fantastisch: An INTERVAL represents a duration, so one could represent it as number, but this would mean one must assume a unit (eg milliseconds). But maybe string is more suitable, as thus would prevent such assumptions.

fantastisch:
> Follow-up issue for this change would be to make `test` work with INTERVALS, which is currently broken. :)

Ah, missed that, sorry. I’ll be off for a few days, but can fix that next week!

PR #1381, opened 2026-07-13
export_protobuf add customized package name

Schokuroff:
- [ ] Tests pass (`uv run pytest`)
- [ ] Code formatted (`uv run ruff check --fix && uv run ruff format`)


Schokuroff:
take a look pls @jschoedl

tests/test_export_protobuf.py
```diff
@@ -19,6 +19,10 @@ def test_to_protobuf():
 kind: DataContract
 apiVersion: v3.1.0
 id: test_protobuf
+description:
+  customProperties:
+    - property: proto_package_name
+      value: com.example.product
```
jschoedl: - Why a `description.customProperty` instead of a `customProperty` at the top level?
- In other cases, we used camelCase for naming custom properties
- As this is no official syntax, it should be explained in the docs

jschoedl (changes requested):
Thanks, makes sense to change that.

Schokuroff:
@jschoedl fixed

tests/test_export_protobuf.py
Schokuroff: fixed

Schokuroff:
@jschoedl take a look plz

PR #1390, opened 2026-07-22
added Optional for message type

Schokuroff:
I think we briefly discussed whether the message type field should be optional before, but I want to discuss it again. Plus, I've added some tests to the previous fix.

- [ ] Tests pass (`uv run pytest`)
- [ ] Code formatted (`uv run ruff check --fix && uv run ruff format`)


jschoedl:
As far as I know, `optional` breaks compatibility with `protoc` versions before 3.15 (February 2021), but its more consistent to the treatment of other field types. I think its fine to include then.

jschoedl:
I removed one of the fixture changes, because the package-name setting is already covered in the tests.

PR #1406, opened 2026-07-26
fix(test): compare structured physicalType fields by name

### LINKED ISSUE #1377: physicalType check fails on Snowflake for type aliases and structured types
`datacontract test` fails the `physicalType` check for Snowflake columns whose declared type is spelled differently than `INFORMATION_SCHEMA` reports it, and for structured `OBJECT(...)` columns.

```python
from datacontract.engines.checks.physical_type_match import physical_type_matches as m
from datacontract.engines.ibis.native_type import reconstruct_native_type as r

# Snowflake INFORMATION_SCHEMA rows for: id INT, name TEXT
m("INT", r("NUMBER", None, 38, 0))             # -> (False, "expected physical type 'INT' but the column is 'NUMBER(38)'")
m("NUMBER", r("NUMBER", None, 38, 0))          # -> (False, ...)  # bare expected should match any params
m("VARCHAR", r("TEXT", 16777216, None, None))  # -> (False, ...)
```

Expected: all three pass. `INT` and `NUMBER` are the same Snowflake type, as are `VARCHAR` and `TEXT`, and per `physical_type_match.py`'s own docstring a bare declared type matches a parameterized column.

Two causes:

1. sqlglot's Snowflake dialect expands bare `NUMBER` to `DECIMAL(38, 0)`, so `_params(exp_dt)` is non-empty and gets compared even though the contract declared no parameters.
2. `reconstruct_native_type` omits a zero scale, producing `NUMBER(38)` where the parsed expected side carries `['38', '0']`.

Separately, structured types never match, because `INFORMATION_SCHEMA.COLUMNS.data_type` strips the field list (`OBJECT(a INT, b TEXT)` is reported as bare `OBJECT`), and because `_params` compares rendered strings rather than recursing with the same alias-aware rule:

```
'OBJECT(a INT, b TEXT)' vs 'OBJECT(a NUMBER(38,0), b VARCHAR(16777216))'  -> False
'OBJECT(b TEXT, a INT)' vs 'OBJECT(a INT, b TEXT)'                        -> False  (field order)
```

The structured actual type is available from `SHOW COLUMNS` (already read by `snowflake_structured_types.py`).

Sanjays2402:
Closes #1377

The `physicalType` check compared a structured type's rendered parameter list, so a Snowflake `OBJECT` column never matched its own contract: `OBJECT(a INT, b TEXT)` was compared string-wise against the catalog's `OBJECT(a NUMBER(38,0), b VARCHAR(16777216))`, field order mattered, and a bare `OBJECT` (what `INFORMATION_SCHEMA` reports when the field list is unavailable) failed too. Structured types are now split into named fields and compared field by field, order-independently, reusing the same alias-aware base-type rule already used for top-level types. Genuine differences still fail.

The three scalar cases in the issue (`INT`/`NUMBER`/`VARCHAR` vs the reconstructed Snowflake types) already pass on `main`, so this PR addresses the remaining structured-type half.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Fixed
 - BigQuery export failed on fields with `logicalType: time`
 - Testing Parquet files failed for `number` fields without a declared precision and scale
+- `datacontract test` no longer fails the `physicalType` check for structured types such as Snowflake `OBJECT(a INT, b TEXT)`, whose fields are now compared by name with the same alias-aware rule as top-level types (#1377)
 
 ## [1.0.14] - 2026-07-23
 
--- a/datacontract/engines/checks/physical_type_match.py
+++ b/datacontract/engines/checks/physical_type_match.py
@@ -87,6 +87,17 @@ def _params(dtype: exp.DataType) -> list[str]:
     return [e.sql().strip().lower() for e in dtype.expressions]
 
 
+def _fields(dtype: exp.DataType) -> Optional[list[Tuple[str, exp.DataType]]]:
+    """Named fields of a structured type (``OBJECT(a INT, b TEXT)``), or ``None``.
+
+    Returns ``None`` for ordinary parameterized types, whose ``expressions`` are
+    length/precision literals rather than field definitions.
+    """
+    if not dtype.expressions or not all(isinstance(e, exp.ColumnDef) for e in dtype.expressions):
+        return None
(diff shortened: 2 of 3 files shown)


jochenchrist:
Thanks for the submission. I am closing this PR because it appears to be part of a large-scale automated AI contribution pipeline: this account has opened roughly 100 similar pull requests across unrelated open-source projects within the past week. We do not accept unsolicited PRs generated this way, as the review and maintenance burden outweighs the benefit.

Issue #1377 remains open and will be addressed separately.

PR #1407, opened 2026-07-26
Fix SQL server schema

j-w-brown:
Fix:
Update ibis connection method to include the specified database schema when using SQL Server.


simonharrer:
Duplicate of #1402 as mssql and sqlserver are synonyms.

PR #1454, opened 2026-07-29
Run individual quality rules with test --quality-id and --tag

### LINKED ISSUE #1080: run a single test from datacontract.yaml
Hi Team

**Our Use Case**
We have a datacontract.yaml file for a table that includes several quality tests. These tests are critical, non-critical, performance-intensive, or less performance-intensive.
We would like to execute these tests based on their category or run them in parallel (at different places or times in our data pipeline) without splitting the main datacontract.yaml into multiple files.
Currently, I see that the datacontract test option provides granularity at the YAML file level, but not at the level of tags or individual quality test names.

**Requested Feature**
It would be very useful for us if we could assign tags to our tests and run them selectively. For example, using a command like this: `datacontract test --server production datacontract.yaml --tag [my_tag]`
Are there any plans to support such functionality in the future?

Thanks in advance for your response and assistance!

simonharrer:
Closes #1080.

## Problem

`datacontract test` selects checks by category (`--checks`) or by dimension (`--dimension`). Neither addresses an individual rule, so running some of a table's rules on ingest and the rest after the nightly load meant splitting the contract into several files — which is exactly what the reporter did not want to do.

## Solution

ODCS quality rules already have an `id` and `tags`. Both were parsed and then dropped on the floor. They now travel with the check from the contract through the check IR into the run, and `test` can filter on them:

```bash
datacontract test --quality-id order_id_is_unique datacontract.yaml
datacontract test --tag critical datacontract.yaml
datacontract test --tag critical,cheap --server production datacontract.yaml
```

```yaml
schema:
  - name: orders
    quality:
      - id: orders_not_empty
        type: library
        metric: rowCount
        mustBeGreaterThan: 0
        tags: ["critical", "cheap"]
```

Behaviour worth reviewing:

- **Quality only.** Both options select quality rules and nothing else — no schema or service level checks, and the JSON Schema validation is skipped for the same reason. Only a rule carries an id or a tag, so anything else would be arbitrary.
- **An unknown `--quality-id` fails the run**, listing the ids the contract does declare. An id is a precise reference like `--schema-name`, and a typo that silently tests nothing would report green in a pipeline.
- **An unmatched `--tag` reports no checks**, the way `--checks` and `--dimension` already behave for a selector that matches nothing.
- **`--tag` matches the rule's own `tags`**, not the tags of a schema or property. Selecting every check of a `pii`-tagged column is a plausible next step, but it is a different feature and would make `--tag` mean two things.
- Filters combine: `--tag critical --checks quality` is an AND, consistent with `--checks` + `--dimension` today.
- Test results now report the `quality_id` and `tags` of the rule each check came from, so a failing check names the rule to re-run.

## Notes

The Azure blob engine emits checks as it goes instead of building a `CheckSpec` list, so it filters per check. Rather than thread a fourth pair of parameters through six helpers, its criteria are bundled into a `CheckFilter` (`datacontract/engines/checks/check_filter.py`). The public `check_azure_blob_file(...)` keyword arguments are unchanged.

## Tests

- `tests/test_test_quality_id_and_tag_filter.py` — 14 new tests: selection by id and tag, multiple values, combination with each other, unknown id failing the run, rules without an id/tags never being matched, ids and tags reported on the results, and the CLI options end to end.
- `tests/test_test_azure_blob_file.py` — 3 new tests covering the same filters on the Azure blob path.
- Full suite green locally (1242 passed, 7 skipped; excludes the container-backed backends).

## Docs

- `docs/docs/quality-rules/index.md` — new "Identifying rules" section and the two new options under "Running only quality checks".
- `docs/docs/testing/index.md` — the options overview.
- `docs/docs/commands/test.md` — regenerated with `update_command_docs.py`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## [Unreleased]
 
 ### Added
+- `datacontract test --quality-id` runs a single quality rule by its ODCS `quality.id`, and `--tag` runs every quality rule declaring one of the given `quality.tags` (#1080)
+- Test results report the `quality_id` and `tags` of the quality rule a check comes from
 - New `databricks-runtime` extra for installing inside a Databricks Runtime, where the cluster already provides PySpark: `pip install datacontract-cli[databricks-runtime]` (#1211 @chifu1234)
 - The docs Commands reference documents the global options `--version` and `--system-truststore`, and every import and export guide links to its command page and back
 - `datacontract test --dimension` runs only the checks measuring one data quality dimension, e.g. `--dimension uniqueness`; it matches the ODCS `quality.dimension` of a rule and the schema and service level checks that measure the same aspect
--- a/datacontract/command_test.py
+++ b/datacontract/command_test.py
@@ -56,6 +56,17 @@ def _parse_enum_csv(value: str | None, enum_cls: type[Enum], option: str, label:
     return {enum_cls(v).value for v in raw}
 
 
+def _parse_csv(value: str | None, option: str) -> set[str] | None:
+    """Parse a comma-separated option into a set of values, or None if unset."""
+    if value is None:
+        return None
+    values = {v.strip() for v in value.split(",") if v.strip()}
+    if not values:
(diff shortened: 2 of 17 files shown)


BennoGBN:
Great, many thanks!

PR #1457, opened 2026-07-30
Support programmatic Databricks and Snowflake credentials

jschoedl:
Adds programmatic credentials for Databricks and Snowflake using Pydantic Settings, so library users embedding the CLI no longer have to mutate `os.environ` per request. Draft for #1250, possible alternative to #1339.

```python
DataContract(
    data_contract_file="datacontract.yaml",
    source_config=DatabricksSourceConfig(token=tenant_token, http_path=tenant_path),
).test()
```

Fields left unset fall back to their environment variable individually, so existing environment-only setups are unchanged. Also fixes a `TypeError` when `DATACONTRACT_SNOWFLAKE_ACCOUNT`, `..._DATABASE`, or `..._SCHEMA` was set alongside the contract's own values.

Adds `pydantic-settings` as a dependency; its own requirements are already in the tree.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -8,3 +8,19 @@ repos:
       args: [ --fix ]
     # Run the formatter.
     - id: ruff-format
+
+# Regenerate the committed generated docs, so CI never fails on stale ones.
+- repo: local
+  hooks:
+    - id: command-docs
+      name: Regenerate docs/docs/commands from the CLI --help output
+      entry: python update_command_docs.py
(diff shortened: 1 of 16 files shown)


jochenchrist:
Implemented in #1467: credentials and connection options (Databricks, Snowflake, and all other sources) can be passed programmatically via DataContract(config=Config(...)), a dict, a YAML config file (--config-file), or per-request datacontract-* headers on the API server. See the new Configuration docs page for details.

PR #1466, opened 2026-08-01
Add row-level filtering to `datacontract test` via --filter and --filters

### LINKED ISSUE #1463: Add row-level filtering / where clause to `datacontract test`
# Feature description
When running `datacontract test` on large datasets (e.g., historical tables with millions/billions of rows in Snowflake, BigQuery, or Databricks), scanning the entire table for data quality and schema checks is extremely expensive and slow. 

In many production pipelines, we only want to validate the incremental data (e.g., data ingested in the last 24 hours or the latest batch).

#### Describe the solution you'd like
We would like to introduce an optional parameter to `datacontract test` that filters the rows being evaluated by applying a condition/WHERE clause to the underlying test query.

Proposed parameter options:
* `--where` clause (Recommended, standard in data tools like dbt)
* `--row-filter` clause

# Example Usage
```bash
# Test only rows ingested in the last 24 hours
datacontract test datacontract.yaml --server production --where "ingested_at >= CURRENT_TIMESTAMP() - INTERVAL '24 hours'"

# Test a specific partition/batch
datacontract test datacontract.yaml --server production --where "batch_id = '2026-07-31'"
```

# Describe alternatives you've considered
Creating temporary views in the data warehouse with the filter applied before running datacontract test, but this adds unwanted DDL overhead and infrastructure complexity to simple test pipelines.

# Additional context
This feature would significantly lower compute costs for large-scale data quality verification and speed up CI/CD or post-ingestion pipeline runs.

jochenchrist:
Closes #1463.

Adds row-level filtering to `datacontract test`, so large tables can be validated incrementally (last 24 hours, one partition) instead of full-scanned:

```bash
# Single-schema contract (or combined with --schema-name)
datacontract test datacontract.yaml --server production --filter "ingested_at >= CURRENT_TIMESTAMP - INTERVAL '24 hours'"

# Per-schema predicates for contracts with multiple schemas, as a JSON object
datacontract test datacontract.yaml --server production --filters '{"orders": "batch_id = '"'"'2026-07-31'"'"'"}'
```

## Design

This follows the patterns of other data quality engines (dbt `where` config, Soda dataset filters, GX row conditions, Databricks DQX per-check `filter`), addressing the two concerns raised in the issue:

- **Which table does the filter apply to?** Predicates are scoped per schema. `--filters` takes a JSON object mapping schema name to predicate; the `--filter` shorthand takes a bare predicate and is accepted when exactly one schema is under test (single-schema contract or `--schema-name`), erroring otherwise instead of silently applying a clause to tables where the column may not exist.
- **Results must state they cover only a slice.** The applied filters are recorded on the `Run` as a new `filters` field (keyed by schema name), flowing into the JSON output and the published results. The console prints a `Row filter: <schema> WHERE <predicate>` line, and since the filter is applied to the ibis table expression, the recorded per-check SQL (`implementation`) contains the `WHERE` clause.

## Implementation

The filter is applied once per model in the ibis engine, right after table resolution: the bound table is wrapped via `Table.alias(...).sql("SELECT * FROM <alias> WHERE <predicate>")`, so the predicate is passed through in the server's native dialect and every downstream query picks it up automatically: the batched count aggregation, duplicates, freshness/retention, and failed-row samples.

Not filtered (documented on the testing page): schema checks (catalog introspection, no rows involved), custom SQL quality checks (`quality.type: sql` runs the author's query as-is), and JSON schema file validation.

## Testing

- 13 tests in `tests/test_test_row_filter.py` covering the library API and CLI: filtered vs. unfiltered results, `Run.filters` recording, WHERE in the recorded SQL, multi-schema ambiguity, unknown schema names, invalid JSON, conflicting options.
- Related suites pass (checks/dimension/quality-id filters, diagnostics, quality, output formats, failed samples, docs consistency).


pierre-monnet (approved):
LGTM

PR #1481, opened 2026-08-04
Sign released Docker images with cosign

simonharrer:
Adds cosign keyless signing to the released Docker images, on both Docker Hub and the Amazon ECR Public mirror.

The PyPI artifacts are already signed with Sigstore, and the images already carry an SBOM and SLSA provenance (`sbom: true`, `provenance: mode=max`). Signatures were the missing piece: nothing tied the published image to this release workflow in a way a consumer could verify. This also unblocks users running admission policies (Kyverno, Gatekeeper) that require a verifiable signature.

## Changes

**`.github/workflows/release.yaml`**

- `docker` job gets `id-token: write` and signs the image with `cosign sign` after the push. Keyless — the signing identity is the workflow's GitHub OIDC token, so there is no key material to manage or rotate.
- Signs the **digest** (`datacontract/cli@sha256:…`) rather than a tag. Tags are mutable, and the digest covers the whole multi-arch index.
- `push-to-ecr` job switches from `docker buildx imagetools create` to `cosign copy`. `imagetools create` copies the manifest only and would leave the signature (a separate `sha256-<digest>.sig` tag) and the attestations behind on Docker Hub; `cosign copy` carries all three across.

**Docs** — a "Verifying the image" section in `docs/docs/installation.md`, a pointer to it from the README, and a CHANGELOG entry. The section also documents the ECR Public mirror, which was previously published but undocumented.

## Verification

```bash
cosign verify datacontract/cli:latest \
  --certificate-identity-regexp '^https://github\.com/datacontract/datacontract-cli/\.github/workflows/release\.yaml@refs/tags/v' \
  --certificate-oidc-issuer https://token.actions.githubusercontent.com
```

Both flags are required and documented as such — without pinning the identity and issuer, `cosign verify` accepts any valid Sigstore signature, including one produced by someone else.

## Notes

- Signing steps are gated on `github.event_name != 'pull_request'`, matching the existing `push:` condition on the build step.
- The signing itself can only be exercised on a real tag push, since it needs the OIDC token and a pushed digest. The workflow YAML parses and the docs site builds.
- The docs claim signatures start at 1.1.0, matching the current `Unreleased` target in the CHANGELOG. Worth a second look if that target moves.

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -164,6 +164,9 @@ jobs:
     needs:
       - test
       - check_version
+    permissions:
+      contents: read
+      id-token: write  # IMPORTANT: mandatory for keyless signing with cosign
     steps:
       - name: Checkout
         uses: actions/checkout@v4
@@ -195,6 +198,7 @@ jobs:
           username: ${{ secrets.DOCKERHUB_USERNAME }}
           password: ${{ secrets.DOCKERHUB_TOKEN }}
       - name: Build and push
+        id: build
         uses: docker/build-push-action@v5
         with:
           context: .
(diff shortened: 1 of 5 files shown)


.github/workflows/release.yaml
```diff
@@ -207,14 +211,23 @@ jobs:
           labels: ${{ steps.meta.outputs.labels }}
           sbom: true
           provenance: mode=max
+      - name: Install cosign
+        if: ${{ github.event_name != 'pull_request' }}
+        uses: sigstore/cosign-installer@v3
```
mahic: Suggest using the latest version of the action, preferably pinned. Cosign releases are not yet made immutable and the [Trivy hack](https://github.com/aquasecurity/trivy/security/advisories/GHSA-69fq-xp46-6x23)  shows how prone to attacks the action supply chain is.

simonharrer: thanks. Fixed this.

mahic:
@simonharrer why did you revert the change from pinned sha?

PR #1482, opened 2026-08-04
feat: add breaking change detection

### LINKED ISSUE #1016: breaking, diff, changelog for ODCS
bring these commands back and adopt for ODCS:

- breaking
- changelog
- diff

### LINKED ISSUE #1129: Reintroduce `breaking`
I'd like to reintroduce breaking change detection, which was removed during the ODCS migration (ref #1016), and wanted to align on an approach before starting.

My current thinking:

  - The existing DCS-based rules ([breaking_rules.py](https://github.com/datacontract/datacontract-cli/blob/v0.10.41/datacontract/breaking/breaking_rules.py) @ v0.10.41) would serve as the baseline and be translated to equivalent ODCS schema rules.
  - ODCS fields that have no DCS counterpart would be evaluated independently for whether they warrant new rules.

Before proceeding, I wanted to check: have there been any prior discussions or design decisions on this topic that I should be aware of? Any pointers to preferred approach or scope would be appreciated.

Thanks,

cc @jochenchrist @jschoedl

pierre-monnet:
Closes https://github.com/datacontract/datacontract-cli/issues/1016


pierre-monnet:
@jochenchrist @simonharrer

datacontract/breaking/rules.py
```diff
@@ -0,0 +1,235 @@
...
+    def evaluate(self, entry: ChangelogEntry) -> RuleEvaluation:
+        return RuleEvaluation(self.rule_id, BreakingChangeLevel.INFO, _change_message("contract", entry))
+
+
+def _is_schema_property_path(segments: list[str]) -> bool:
+    try:
+        properties_index = segments.index("properties")
+    except ValueError:
+        return False
+    return segments[0] == "schema" and properties_index + 1 < len(segments)
```
jschoedl: This returns `True` if any of the segments is `properties`, even if it is not the last one. E.g. deleting a description (located at `properties.some_property.description`) is mis-classified as deleting the property itself, although this is actually no breaking change.

We should check for `properties` being the segment before the last one

docs/docs/testing/breaking-changes.md
jschoedl, line 1: This file should not be placed in the `testing` directory, because it hides the `datacontract breaking` functionality between the list of different testing backends.

Instead, I'd vote for a seperate page "Compare contract versions" (or similar name) at the top level, which should include both `changelog` and `breaking`. As there is no seperate documentation of `changelog` yet, you don't need to write it - but at least a link to the auto-generated command reference would be nice.

datacontract/breaking/detector.py
```diff
@@ -0,0 +1,76 @@
...
+
+    def __init__(self, rules: Iterable[BreakingChangeRule] = DEFAULT_RULES):
+        self._rules = tuple(rules)
+        if not self._rules:
+            raise ValueError("At least one breaking-change rule is required")
+
+    def detect(self, changelog: ChangelogResult) -> BreakingChangeResult:
+        entries = [self._classify(entry) for entry in changelog.entries]
+        summary = [self._summarize(entry, entries) for entry in changelog.summary]
+        return BreakingChangeResult(v1=changelog.v1, v2=changelog.v2, summary=summary, entries=entries)
```
jschoedl, line 38: `_summarize(entry, entries)` loops each of the changelog entries, i.e. we have quadratic runtime complexity. If there are a lot of entries (e.g. a few thousands), this can take a while.

To avoid that, the `entries` parameter should not get all of the entries, but only the relevant ones, e.g. like this:

```suggestion
    def detect(self, changelog: ChangelogResult) -> BreakingChangeResult:
        entries = [self._classify(entry) for entry in changelog.entries]
        entries_by_prefix = defaultdict(list)
        for entry in entries:
            prefix = ""
            for segment in entry.path.split("."):
                prefix = f"{prefix}.{segment}" if prefix else segment
                entries_by_prefix[prefix].append(entry)
        summary = [self._summarize(entry, entries_by_prefix.get(entry.path, [])) for entry in changelog.summary]
        return BreakingChangeResult(v1=changelog.v1, v2=changelog.v2, summary=summary, entries=entries)
```

datacontract/api.py
```diff
@@ -863,6 +863,42 @@ async def changelog_endpoint(
...
+@app.post(
+    "/breaking",
+    tags=["breaking"],
+    summary="Show compatibility impact between two data contracts.",
+    description="""
+        Compare two ODCS data contract YAMLs and classify their backward-compatibility impact.
+        POST a JSON body with `v1` (source/before) and `v2` (target/after) as YAML strings.
+    """,
+)
+async def breaking_endpoint(
```
jschoedl:
```suggestion
def breaking_endpoint(
```

This is declared as `async`, but actually only does blocking CPU and I/O.

tests/test_breaking.py
```diff
@@ -0,0 +1,104 @@
...
+from datacontract.model.changelog import ChangelogEntry, ChangelogResult, ChangelogType
+
+V1 = "fixtures/changelog/integration/changelog_integration_v1.yaml"
+V2 = "fixtures/changelog/integration/changelog_integration_v2.yaml"
+
+
+def _entry(path, change_type, old_value=None, new_value=None):
+    return ChangelogEntry(path=path, type=change_type, old_value=old_value, new_value=new_value)
+
+
```
jochenchrist, line 40: When looking at the tests, I cannot really see how the command works from a CLI perspective?

pierre-monnet: Added some tests. Let me know if more are needed!

datacontract/breaking/rules.py
```diff
@@ -0,0 +1,235 @@
...
+        if old is not None and new is not None:
+            if set(old) - set(new):
+                level = BreakingChangeLevel.ERROR
+            elif set(new) - set(old):
+                level = BreakingChangeLevel.INFO
+            else:
+                level = BreakingChangeLevel.INFO
+        else:
+            level = BreakingChangeLevel.WARNING
+        return RuleEvaluation(self.rule_id, level, _change_message("enum constraint", entry))
```
jschoedl: `enum` is not supported in ODCS. As it was supported in the old DCS format, there might be some locations in code or docs where we didn't clean it up yet, but today, this code is not reachable.

The new way is to specify a quality check with `validValues`: https://bitol-io.github.io/open-data-contract-standard/v3.1.0/data-quality/#invalid-values

datacontract/breaking/rules.py
```diff
@@ -0,0 +1,235 @@
...
+        if not entry.path.endswith(self._suffixes):
+            return None
+        old = _parse_number(entry.old_value)
+        new = _parse_number(entry.new_value)
+        if old is None or new is None:
+            level = BreakingChangeLevel.WARNING
+        elif _is_tightening(entry.path, old, new):
+            level = BreakingChangeLevel.ERROR
+        else:
+            level = BreakingChangeLevel.INFO
```
jschoedl, line 132: `_parse_number` in its current form does not work for `logicalType: date` yet

datacontract/api.py
```diff
@@ -863,6 +863,42 @@ async def changelog_endpoint(
...
+    with tempfile.NamedTemporaryFile(mode="w", suffix=".yaml", delete=False) as f1:
+        f1.write(body.v1)
+        v1_path = f1.name
+    with tempfile.NamedTemporaryFile(mode="w", suffix=".yaml", delete=False) as f2:
+        f2.write(body.v2)
+        v2_path = f2.name
+
+    try:
+        result = DataContract(data_contract_file=v1_path).breaking(DataContract(data_contract_file=v2_path))
+        return result
```
jschoedl: This includes internal information that is not needed by the caller such as the full file paths of the tempdir. For security reasons, it would be best to keep the response minimal as it was done with the `ChangelogResponse` a few lines before this one.

datacontract/api.py
```diff
@@ -863,6 +863,42 @@ async def changelog_endpoint(
...
 
+@app.post(
+    "/breaking",
+    tags=["breaking"],
+    summary="Show compatibility impact between two data contracts.",
+    description="""
+        Compare two ODCS data contract YAMLs and classify their backward-compatibility impact.
+        POST a JSON body with `v1` (source/before) and `v2` (target/after) as YAML strings.
+    """,
+)
```
jschoedl, line 982: Please add `operation_id`, `response_description` and `responses` to document the API endpoint (see the other endpoints for examples)

datacontract/breaking/rules.py
jschoedl, line 1: The `priority` integer seems a bit odd to me. Why not list all the rules in a list where the first one wins?

jschoedl (commented):
Hi, thank you for the PR!

pierre-monnet:
> Hi, thank you for the PR!

Hi, thanks for review, I pushed fixes :)

jschoedl:
Hey @pierre-monnet, thanks again for picking this up. I now looked at your PR in detail again. This is a rather complex feature with many small detail decisions - to avoid too many review rounds, I just made the changes that I'd like to add. Please object if there is any change you wouldn't agree on!
Otherwise I'll merge it soon so it gets into the next release.

PR #1489, opened 2026-08-06
Resolve physicalName in servicelevel freshness and retention checks

### LINKED ISSUE #1488: `datacontract test` freshness/retention checks ignore `physicalName` and read the logical schema object name
`datacontract test` resolves every schema check through the schema object's `physicalName`, but the freshness and retention checks read the relation by the logical `name`: the servicelevel check builders split `sla.element` directly and skip the `to_schema_name()` resolution every schema check uses. Any contract where the two names differ — the normal case for a versioned dbt model, which dbt aliases to `<model>_v<n>` — fails with `Could not read model '<name>'` while all schema and quality checks in the same run pass against the correct relation. `datacontract export sodacl` has the same defect: the freshness/retention blocks are emitted under `checks for <logical name>`.

Steps to reproduce

*Step 1* — a table whose name differs from the contract's schema object name:

```bash
docker run -d --rm --name repro -e POSTGRES_USER=postgres -e POSTGRES_PASSWORD=postgres -p 5433:5432 postgres:16
docker exec repro psql -U postgres -c "CREATE TABLE public.events_v1 (ts TIMESTAMPTZ); INSERT INTO public.events_v1 VALUES ('2023-01-01');"
```

*Step 2* — `freshness-repro.odcs.yaml`:

```yaml
apiVersion: v3.1.0
kind: DataContract
id: freshness-repro
name: freshness-repro
version: 0.0.1
status: active
servers:
- server: postgres
  type: postgres
  host: localhost
  port: 5433
  database: postgres
  schema: public
schema:
- name: events
  physicalName: events_v1
  logicalType: object
  physicalType: table
  properties:
  - name: ts
    logicalType: timestamp
slaProperties:
- property: freshness
  value: 1
  unit: h
  element: events.ts
```

*Step 3*

```bash
export DATACONTRACT_POSTGRES_USERNAME=postgres DATACONTRACT_POSTGRES_PASSWORD=postgres
datacontract test freshness-repro.odcs.yaml --checks servicelevel
```

Results in following output

```
╭────────┬─────────────────────────────┬───────┬───────────────────────────────╮
│ Result │ Check                       │ Field │ Details                       │
├────────┼─────────────────────────────┼───────┼───────────────────────────────┤
│ failed │ Freshness of events.ts < 1h │ ts    │ Could not read model          │
│        │                             │       │ 'events': events              │
╰────────┴─────────────────────────────┴───────┴───────────────────────────────╯
```

Expected: the check reads `events_v1` — like the schema checks in the same run do — and measures freshness. Renaming the schema object to `events_v1` (and the element to `events_v1.ts`) measures correctly, which shows only the name resolution is at fault.

Reproduced on 1.0.15, 1.0.17, 1.1.0, and current `main`; present since executable freshness/retention slaProperties were introduced in 0.11.1.

Happy to submit a PR — fix is ready.

erikgrip2:
Fixes #1488.

`datacontract test`'s freshness and retention checks read the relation by the schema object's logical `name`; every other check resolves `physicalName` via `to_schema_name()`. A contract whose object name differs from the relation name — e.g. a versioned dbt model, aliased to `<model>_v<n>` — failed while all schema checks in the same run passed against the correct relation:

```
│ failed │ Freshness of events.ts < 1h │ ts │ Could not read model 'events': events │
```

`datacontract export sodacl` emitted its servicelevel blocks under the logical name the same way.

## Changes

- Engine: a `_resolve_sla_element` helper folds the element split, the model existence gate, and name resolution — the object via `to_schema_name()` (keeping the kafka carve-out, which reads the Spark SQL view by logical name), the property via its `physicalName`. Both `_freshness_check` and `_retention_check` use it.
- Sodacl export: a `_resolve_physical_names` helper of the same shape, used by both servicelevel builders.
- The postgres servicelevels fixture now has an object name that differs from `physicalName`, so the e2e can distinguish "measured and stale" from "could not read the table"; it now also asserts the freshness diagnostics (`age_seconds`).
- New unit tests for both layers: resolution, fallback without physicalName, kafka.

## Behavior notes

1. Reported names become physical for servicelevel checks (`Check.model` and the check display name) — consistent with what schema checks already report, but different from the `sla.element` spelling.
2. Row filters (`--filter`/`--filters`) now apply to servicelevel checks when `physicalName` is set. Filters are normalized into a mapping keyed by physical name, and the servicelevel specs' logical-name key never matched, so they silently ran unfiltered.
3. If the logical name happened to also name a readable relation, freshness previously measured that (wrong) table silently; it now measures the right one.
4. File servers (local/s3/gcs/azure): the duckdb views are registered under the logical name, so a contract with a differing `physicalName` already fails every schema check there; servicelevel checks now fail consistently with them instead of accidentally measuring the logical-name view. (Whether those views should be created via `to_schema_name()` instead may deserve a separate look.)

No change for contracts where `physicalName` is unset or equals the name.

## Testing

- New unit tests fail before the fix and pass after.
- The postgres e2e fails pre-fix with `Could not read model '<logical name>'` and passes post-fix with freshness measurement diagnostics.

---


datacontract/export/sodacl_check_builder.py
jschoedl: Now that `field_name` is the `physicalName`, please wrap it with `_quote_identifier_if_needed(...)`.

jschoedl:
Thank you, LGTM!

PR #1492, opened 2026-08-08
Add `datacontract import pydantic-model`

OGsiji:
`datacontract export pydantic-model` exists, but there is no way back. This adds the importer.

```bash
datacontract import pydantic-model --source models.py
```

The module is parsed with `ast` rather than imported. Deriving a contract from a service's `models.py` shouldn't require installing that service's dependencies, and shouldn't run its import-time side effects — which is the case I care about, pointing the CLI at an API repo from CI.

How the models map:

- Module-level models become schema objects. Models used as the type of another model's field are inlined as nested objects instead of becoming schemas of their own.
- `Optional[X]`, `X | None`, and a default value set `required`.
- `Field(...)` and `Annotated[..., Field(...)]` constraints become `logicalTypeOptions`, and `description` becomes the property description.
- `Literal` types and `Enum` classes become an `invalidValues` quality rule, the same shape the JSON Schema importer emits.
- Nested models, `list[...]` and `dict[...]` map to object, array and object.

No new dependency — `ast` is stdlib.

There's a round-trip test (import → `export pydantic-model` → import), and that's what turned up the first commit. An object property with no sub-properties made the exporter emit

```python
class Metadata(pydantic.BaseModel):
```

with an empty body, so the generated module didn't parse. The fix is a separate commit with its own test in `test_export_pydantic.py`, in case you'd rather take it on its own.

Docs page, example under `examples/imports/pydantic-model/`, and CHANGELOG are updated; `update_import_examples.py` and `update_command_docs.py` re-run.

One thing I wasn't sure about: I mapped `bytes` to logical type `array` to match the Avro and SQL importers, but that's the one column the round trip can't preserve, because the exporter writes an item-less array as `list[typing.Any]`. Happy to map it to `string` instead if you'd rather it round-trip.


siji-circo:
@jochenchrist  Please can you look into this PR. , let me know if it is helpful or I DRAFT it

simonharrer:
Thanks for your contribution. This is great. I've merged this. If you want, add an additional rountrip test so we really know that the export and the import work perfect together, in an additional PR.

OGsiji:
Hey @simonharrer , thanks for your prompt response, I will create the other PR
Thanks.

PR #1497, opened 2026-08-12
Add dataset, project, catalog, and database placeholders to SQL quality rules

aabf:
Follows up on the [Slack thread in #datacontract](https://datacontract.slack.com/archives/C06L4GBKJC8/p1786378830588679) about `type: sql` quality rules against BigQuery.

## The gap

`prepare_query()` can substitute the model name (`${model}` / `${table}` / `${object}`) and the server's schema (`${schema}`), and nothing else. BigQuery has no schema concept - an ODCS BigQuery server sets `project` and `dataset` - so `schema` is never set there and:

- `SELECT COUNT(*) FROM ${table}` fails with `400 Table "my_table" must be qualified with a dataset (e.g. dataset.table)`
- `${schema}.${table}` resolves to `my_table.my_table`, because `${schema}` falls back to the model name

That leaves no way to write a rule against BigQuery without hardcoding `project.dataset.table` into every query, which defeats the point of naming the location once in the server block.

## The change

`${dataset}`, `${project}`, `${catalog}`, and `${database}` now resolve to the matching value on the selected server, so a rule can be written as:

```yaml
quality:
  - type: sql
    query: SELECT COUNT(*) FROM ${project}.${dataset}.${table}
    mustBeGreaterThan: 0
```

Applied in both query preparation paths - `datacontract/engines/checks/create_checks.py` (ibis) and `datacontract/export/sodacl_check_builder.py` (sodacl, where the new placeholders get the same identifier quoting as `${model}` and `${schema}`). Like `${schema}`, a placeholder the server has no value for falls back to the model name.

The `if/elif/else` quoting blocks in the sodacl path were folded into one local helper so the new placeholders reuse them rather than repeating the branches a third time.

## One deviation from the thread

The thread also suggested making `${schema}` fall back `schema_ -> dataset -> model_name`. That is not in this PR: with `${dataset}` available, the fallback is a second, implicit path to the same value, and `${schema}` meaning "dataset" on some servers but "schema" on others is harder to document than two distinct placeholders. Happy to add it if you'd rather have both - it's a two-line change.

## Docs and tests

- `docs/docs/quality-rules/sql.md` gains a Placeholders section; the existing placeholders were undocumented, so the table covers all of them, not just the new ones.
- `tests/test_create_checks_sql_placeholders.py` (new) covers the ibis path, and `tests/test_data_contract_checks.py` the sodacl path including backtick quoting.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Added
+- SQL quality rules support the `${dataset}`, `${project}`, `${catalog}`, and `${database}` placeholders for the server's values
+
 ### Changed
 - Test results name the check fields `qualityId` and `failedSamples` instead of `quality_id` and `failed_samples`; the old names are deprecated, but still accepted as input and still written next to the new ones
 - `datacontract import unity` no longer writes the `databricksType` custom property, which duplicated `physicalType`
--- a/datacontract/engines/checks/create_checks.py
+++ b/datacontract/engines/checks/create_checks.py
@@ -141,10 +141,6 @@ def to_threshold(quality: DataQuality) -> Optional[Threshold]:
     return None
 
 
-def _server_value(server: Optional[Server], attribute: str) -> Optional[str]:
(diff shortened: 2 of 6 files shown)


simonharrer:
Great catch. Thanks for the contribution!

PR #1507, opened 2026-08-14
Round trip the Pydantic exporter and importer over every example contract

OGsiji:
The round trip you asked for in #1492, over every contract under `examples/` rather than a fixture I write myself — ten different importers produce those, so they cover column shapes I wouldn't have thought to invent.

It found three things, so the fixes are the first two commits and the test is the third.

**`timestamp` and `time` were never handled.** ODCS v3.1.0 allows `string, date, timestamp, time, number, integer, object, array, boolean`, but the exporter's `match` had no case for `timestamp` or `time`. Both fell through to the physical-type guess, so `physicalType: TIMESTAMPTZ` from the SQL importer came out as a plain `str`, and `TIMESTAMP` from dbml and BigQuery only reached `datetime.datetime` by accident.

Fixing that forced a decision on `date`, which exported as `datetime.datetime`. If `timestamp` also exports as `datetime.datetime` the two are indistinguishable coming back, so `date` now exports as `datetime.date`. That is the one judgement call here — happy to revert it and let the test treat `date` and `timestamp` as equivalent, if you'd rather not change what `date` exports today.

`order_timestamp` in `examples/orders/orders.odcs.yaml` was declared `logicalType: date` with `physicalType: TIMESTAMP` and described as "Timestamp when the order was placed", so I corrected it to `timestamp`. That keeps the Pydantic export identical to before, and it also improves the avro (`int`/`date` → `long`/`timestamp-millis`), dbml, jsonschema (`date` → `date-time`), markdown, mermaid and sqlalchemy (`Column(Date)` → `Column(TIMESTAMP)`) examples generated from that contract.

**The contract description was dropped on import.** The exporter writes it as a string below the imports; the importer ignored it, so it didn't survive. Now both that spelling and a real module docstring are read back.

The test asserts name, logical type, requiredness, description and nesting. It deliberately doesn't assert `physicalType`, constraints, quality rules or examples — a bare annotation can't carry those, and pretending otherwise would just encode today's behaviour. Schema names compare case-insensitively because the exporter names the class after the schema, capitalized (`orders` → `class Orders`).

There's also a fixed-point check: a second lap has to produce the same module as the first. That's what caught the dropped description, since the loss only shows on the way back round.

All 12 examples pass on the first assertion and reach a fixed point.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,11 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Fixed
+- `datacontract export pydantic-model` exports the ODCS `timestamp` and `time` logical types as `datetime.datetime` and `datetime.time` instead of guessing from the physical type, which turned a `TIMESTAMPTZ` column into a `str`
+- `datacontract export pydantic-model` exports the ODCS `date` logical type as `datetime.date` instead of `datetime.datetime`
+- `datacontract import pydantic-model` maps `datetime.datetime` to the `timestamp` logical type and `datetime.time` to `time`
+
 ## [1.1.1] - 2026-08-14
 
 ### Added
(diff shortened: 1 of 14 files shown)


OGsiji:
@simonharrer

simonharrer:
Love it. Thanks!

PR #1512, opened 2026-08-14
Fix S3 JSON validation for multiple files

### LINKED ISSUE #1511: JSON schema validation on S3 silently skips all files but the last
`process_s3_file` (`datacontract/engines/fastjsonschema/check_jsonschema.py:202-217`) validates only the last object matched by the `location` glob. Every other file is downloaded and then discarded unvalidated, so an invalid record in any file but the last goes unreported and the run still appends the passed check "All JSON entries are valid." for the model.

Cause: the loop body only assigns a lazy generator over the current file's content to `json_stream`, overwriting the previous file's generator before anything consumes it; `validate_json_stream` is then called once, after the loop.

Applies whenever `location` matches more than one object — e.g. the documented `s3://my-bucket/orders/*.json`, or a `{model}` pattern expanding onto several files. Secondary effect: the full dataset is downloaded (and paid for in egress) even though only one file is checked.

Fix direction: validate inside the loop and accumulate the exceptions across files.

**Where to start:** the fix is in `process_s3_file` — call `validate_json_stream` per file inside the loop and collect the exceptions from all files before `process_exceptions`.

**How to test:** `tests/test_test_s3_json_multiple_models.py` shows the MinIO/`testcontainers` setup (needs Docker). A regression test uploads two JSON objects matched by one glob, with the invalid record in the *first* one, and asserts the run fails.

nightcityblade:
Fixes #1511

## Summary

- validate every S3 object matched by the configured location while iterating the files
- accumulate schema errors across all matched objects before reporting them
- add a regression test with an invalid first object and valid final object

## Testing

- `.venv/bin/pytest tests/test_test_s3_json_multiple_models.py::test_s3_json_validates_every_matching_file tests/test_test_local_json.py tests/test_test_local_json_nd.py -q` (3 passed)
- `.venv/bin/ruff check datacontract/engines/fastjsonschema/check_jsonschema.py tests/test_test_s3_json_multiple_models.py`
- `.venv/bin/ruff format --check datacontract/engines/fastjsonschema/check_jsonschema.py tests/test_test_s3_json_multiple_models.py`

- [x] Docs updated (N/A; no documentation changes are needed)

### DIFF
--- a/datacontract/engines/fastjsonschema/check_jsonschema.py
+++ b/datacontract/engines/fastjsonschema/check_jsonschema.py
@@ -207,17 +207,20 @@ def process_s3_file(run, server, schema, model_name, validate, config: Config | 
     s3_location = server.location
     if "{model}" in s3_location:
         s3_location = s3_location.format(model=model_name)
-    json_stream = None
+    exceptions: List[DataContractException] = []
+    found_file = False
 
     for file_content in yield_s3_files(s3_endpoint_url, s3_location, config):
+        found_file = True
         if server.delimiter == "new_line":
             json_stream = read_json_lines_content(file_content)
         elif server.delimiter == "array":
             json_stream = read_json_array_content(file_content)
         else:
             json_stream = read_json_file_content(file_content)
+        exceptions.extend(validate_json_stream(schema, model_name, validate, json_stream))
 
-    if json_stream is None:
+    if not found_file:
         raise DataContractException(
             type="schema",
             name="Check that JSON has valid schema",
@@ -225,9 +228,6 @@ def process_s3_file(run, server, schema, model_name, validate, config: Config | 
             reason=f"Cannot find any file in {s3_location}",
             engine="datacontract",
         )
-
-    # Validate the JSON stream and collect exceptions.
-    exceptions = validate_json_stream(schema, model_name, validate, json_stream)
 
     # Handle all errors from schema validation.
     process_exceptions(run, exceptions, config)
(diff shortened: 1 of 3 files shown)


nightcityblade:
Rebased onto the current main branch and resolved the CHANGELOG overlap in 472b5b4. Validation after the rebase: the targeted S3 regression plus local JSON tests pass (3 passed), Ruff check and format checks pass, and git diff --check passes. CI is rerunning on the updated branch.

datacontract/engines/fastjsonschema/check_jsonschema.py
```diff
@@ -207,17 +207,20 @@ def process_s3_file(run, server, schema, model_name, validate, config: Config |
...
 
     for file_content in yield_s3_files(s3_endpoint_url, s3_location, config):
+        found_file = True
         if server.delimiter == "new_line":
             json_stream = read_json_lines_content(file_content)
         elif server.delimiter == "array":
             json_stream = read_json_array_content(file_content)
         else:
             json_stream = read_json_file_content(file_content)
+        exceptions.extend(validate_json_stream(schema, model_name, validate, json_stream))
```
jschoedl, line 226: If there is a maximum of exceptions set, this loop should stop after it was reached so we avoid loading unnecessary data. See the call to `process_exceptions` a few lines below where this is checked.

jschoedl (changes requested):
Hi, thank you for the PR!

datacontract/engines/fastjsonschema/check_jsonschema.py
nightcityblade: Addressed in 7114e7d: the S3 loop now uses the same resolved error limit as process_exceptions and stops before fetching another file once that limit is reached. Added a lazy-generator regression test proving later files are not loaded. Validation: all 3 tests in test_test_s3_json_multiple_models.py passed; Ruff check/format and git diff --check passed.

jschoedl:
I'm closing this PR, as it seems to be part of an LLM-powered code generation pipeline with little human oversight. Such contributions are typically consuming much maintainer time with little gain.

PR #1520, opened 2026-08-18
fix: stop a posted data contract from reaching the machine that runs it

simonharrer:
`POST /test` on `datacontract api` could read and write arbitrary files on the machine running the API — unauthenticated, unless `DATACONTRACT_CLI_API_KEY` is set.

A data contract may declare a `quality.type: sql` rule, whose query is executed verbatim. With `servers[].type: local` the backend is a local DuckDB, so the query runs against the API server's own filesystem rather than a data source the operator configured:

```yaml
servers:
  - server: production
    type: local
    path: /srv/data/orders.json
    format: json
schema:
  - name: orders
    quality:
      - type: sql
        query: "SELECT content FROM read_text('/etc/passwd')"   # returned in the response
        mustBe: 0
      - type: sql
        query: "COPY (SELECT 'x') TO '/home/nonroot/.bashrc'"   # written to disk
        mustBeGreaterThan: -1
```

Arbitrary file write is a path to code execution. `datacontract api --help` warns that tests "may be subject to SQL injection", which reads as *against your data source* — reaching the API host itself is a different asset.

## Three independent controls

**1. Custom SQL must be a read-only query — for every data source.** `quality.type: sql` computes a number from the data, so a query is all it ever needs. Anything else is reported as a failed check and never executed: DDL, DML, `COPY` (a file write on DuckDB and `COPY … TO PROGRAM` on Postgres), `ATTACH`, `INSTALL`/`LOAD`, `SET`, `PRAGMA`, and a second statement smuggled in behind a legitimate one. Validation runs on the query *after* placeholder substitution, uses sqlglot (already a dependency), and fails closed — a query that does not parse is refused rather than passed through.

This one applies to every run, CLI included: a quality rule that drops a table is never legitimate. All 47 quality queries in `tests/fixtures`, `examples` and `docs` pass it.

**2. DuckDB is confined to the contract's own data locations.** For a contract that arrives over HTTP, the connection gets `allowed_paths`/`allowed_directories` set to the locations the contract declares, and `enable_external_access = false`. Applied after extensions are loaded and secrets created, before any contract SQL can run — and never to a connection a Python-library caller passed in.

`lock_configuration` turned out to be both unnecessary and harmful: DuckDB already refuses to re-enable external access or widen the allowed paths once it is off, and locking the configuration stops ibis from setting the session timezone. `test_the_sandbox_cannot_be_undone` asserts the escape attempts stay blocked, so a future DuckDB that stops refusing fails the build rather than quietly making the sandbox decorative.

**3. `servers[].type: local` is refused by the API.** The sandbox confines the contract to the path it *declares* — which does not help when the declared path is `/etc/passwd`. So the API refuses the server type outright.

Deployments that mount their data next to the API serve local files on purpose, so there is an opt-in:

```bash
export DATACONTRACT_CLI_API_ALLOW_LOCAL_FILES=true
```

Deliberately **not** a `Config` field: those are settable per request through a `datacontract-*` header, which would let a caller opt themselves in. `test_a_caller_cannot_opt_themselves_in` covers that.

## Scope

`untrusted_contract` is threaded from `app.state.untrusted_contracts` (which already existed, set only by the API app) through `DataContract` to the engine, so nothing changes for the CLI or the Python library — a file the user names on the command line is still theirs to read. Control 1 is the exception and applies everywhere, by design.

Three existing tests in `tests/test_api.py` used a local fixture as a convenient data source; they now opt in through the new env var.

## What this does not change

An untrusted contract can still run a read-only query against a data source the operator configured. That is what `/test` is for, and it is why the docs tell operators to set `DATACONTRACT_CLI_API_KEY`. Narrowing it further would be a product decision, not a fix.

## Verification

Both original exploits are blocked by default (422) and remain blocked with `DATACONTRACT_CLI_API_ALLOW_LOCAL_FILES=true` (200, but the write is refused by control 1 and the read by control 2).

`tests/test_untrusted_contract.py` adds 30 tests: the read-only classification table, the sandbox escape attempts, a glob location allowing only its directory, the API refusal and its opt-in, and paired assertions that an untrusted contract cannot read another file *while a trusted one still can* — so the negative result cannot pass because of a broken query.

Full suite: 2014 passed. (`test_config_file.py::test_missing_config_file_fails_with_a_clear_message` fails identically on `main` under `pytest -n 8` and passes serially — a pre-existing parallel-mode flake, unrelated.)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - `datacontract api` no longer reloads on file changes by default; pass `--reload` to enable it (development only)
 
 ### Fixed
+- A `quality.type: sql` rule must be a read-only query; DDL, DML, `COPY`, `ATTACH` and the like are reported as a failed check instead of being executed, for every data source
+- `datacontract api` refuses `servers[].type: local`, so a posted data contract cannot read the files of the server running it; set `DATACONTRACT_CLI_API_ALLOW_LOCAL_FILES=true` to allow it
+- `datacontract api` confines the DuckDB connection of a file-based server to the data locations the posted data contract declares
 - The Entropy Data API key is only sent to the Entropy Data host, no longer to any host a data contract URL or `--publish` URL points at; set `ENTROPY_DATA_HOST` for a self-hosted deployment
 - `datacontract api` refuses to send an environment-held data source credential to a host named by the posted contract, preventing credential exfiltration through a crafted `servers` section
 - `datacontract export html` and `datacontract catalog` now HTML-escape data contract field values, closing a stored cross-site scripting hole
--- a/datacontract/api.py
+++ b/datacontract/api.py
@@ -251,6 +251,15 @@ that value, and answers `401` when the header is missing and `403` when it is wr
 Securing the API is strongly recommended: `POST /test` connects to your data sources, and test
 results may contain sensitive information.
 
+## Contracts are untrusted
+
+A posted contract carries SQL and names the hosts to connect to, so the server treats it as
+untrusted input: a `quality.type: sql` rule must be a read-only query, a credential held in the
+server's environment is never sent to a host the contract names, and `servers[].type: local` is
+refused so a caller cannot read the server's own files. Set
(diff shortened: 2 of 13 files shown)


simonharrer:
Squash-merged into main as ce77960.

PR #1523, opened 2026-08-20
Add `datacontract test --dry-run`

### LINKED ISSUE #1510: Dry-run feature for the tests
During dry-run, we only show what we would execute, not actually execute. AI can use that to verify whether the desired checks are actually configured correctly.

OGsiji:
Closes #1510.

```bash
datacontract test datacontract.yaml --dry-run
```
```
⚪ no checks were executed. Planned 33 checks. Took 0.16 seconds.
```

Every check is already registered as a stub before execution begins, so the plan is that list with a result of `skipped`, carrying the assertion each check would have made. The existing output formats render it, so the machine-readable view comes for free:

```json
{ "name": "Check that field order_total has no missing values",
  "implementation": "missing_count(order_total) = 0",
  "result": "skipped" }
```

Since nothing connects, a plan works where a run cannot. On a contract with a Postgres server and no credentials set:

```
$ datacontract test contract.yaml
missing_env_DATACONTRACT_POSTGRES_USERNAME: Required configuration ... is not set.   (exit 1)

$ datacontract test contract.yaml --dry-run
⚪ no checks were executed. Planned 10 checks.                                        (exit 0)
```

### On the two questions from the issue

**Azure blob.** Excluded, as you suggested. Those checks read the file listing to decide which checks exist at all, so there is nothing to plan without reading. A dry run appends a `warning` check saying the plan is incomplete rather than quietly omitting them.

**JSON Schema.** Included, since it looked like the case where an empty plan would be most misleading. The schema is still built and compiled and only the file read is skipped, so a contract that could never validate still fails in a dry run — which seemed closer to the point of the feature than skipping it wholesale.

**`skipped` for the checks** works as you said. The run-level result needed a little more care, though, and it is the one thing I would like a second opinion on.

### The one judgement call

An all-skipped run resolves to `unknown`, which the writer treats as invalid and exits `1`. So a plan initially failed the build, which defeats the CI use case.

My first attempt was to make `calculate_result` report `skipped` when every check is skipped. That broke `test_metadata_only_all_skipped_remains_unknown`, which says plainly that you decided otherwise for `--metadata-only`, so I backed it out.

Instead the run records `dryRun: true` and a dry run resolves to `skipped` on that basis — a plan asserts nothing, so it can neither pass nor fail. `--metadata-only` still resolves to `unknown` exactly as before, and its test is untouched.

The field is also what lets anything reading the JSON tell a plan from a run that happened to skip everything, which matters for the case in the issue: without it, "all checks skipped" is ambiguous.

Happy to change how that is modelled if you would rather it worked differently — it is additive to the `Run` schema, which I did not want to do silently.

### Notes

`datacontract ci` does not get the flag; it seemed odd for a command whose job is to run. Easy to add if you want it.

Full suite green (2103 passed). Nine tests added, the main one asserting the plan lists exactly the checks a real run against the same contract executes, so the plan cannot drift from reality without failing.

### DIFF
--- a/datacontract/command_test.py
+++ b/datacontract/command_test.py
@@ -222,7 +222,7 @@ def test(
         bool,
         typer.Option(
             help="Report the checks that would run, without connecting to the server or reading any data. "
-            "Every reported check has the result 'skipped'."
+            "Reported checks have the result 'skipped', or 'warning' where a check could not be planned."
         ),
     ] = False,
     include_failed_samples: Annotated[
--- a/datacontract/command_test.py
+++ b/datacontract/command_test.py
@@ -218,6 +218,13 @@ def test(
             "Checks that read row values are skipped."
         ),
     ] = False,
+    dry_run: Annotated[
+        bool,
+        typer.Option(
+            help="Report the checks that would run, without connecting to the server or reading any data. "
+            "Every reported check has the result 'skipped'."
+        ),
+    ] = False,
     include_failed_samples: Annotated[
         bool,
         typer.Option(
@@ -284,6 +291,7 @@ def test(
         filter=filter,
         filters=parsed_filters,
         metadata_only=metadata_only,
+        dry_run=dry_run,
     ).test()
     if logs:
         _print_logs(run)
--- a/datacontract/engines/data_contract_test.py
+++ b/datacontract/engines/data_contract_test.py
@@ -117,6 +117,15 @@ def execute_data_contract_test(
             run.log_warn(f"No checks found for tags: {', '.join(sorted(tags))}")
     run.checks.extend(build_check_stubs(specs))
 
+    if metadata_only:
+        executable = []
+        for spec in specs:
+            if spec.requires_data_read:
+                set_result(run, spec.key, ResultEnum.skipped, "Row-value check disabled by --metadata-only")
+            else:
+                executable.append(spec)
+        specs = executable
+
     if dry_run:
         _report_dry_run(
             run,
@@ -131,15 +140,6 @@ def execute_data_contract_test(
             config=config,
         )
         return
-
-    if metadata_only:
-        executable = []
-        for spec in specs:
-            if spec.requires_data_read:
-                set_result(run, spec.key, ResultEnum.skipped, "Row-value check disabled by --metadata-only")
-            else:
-                executable.append(spec)
-        specs = executable
 
     # TODO check server is supported type for nicer error messages
     # TODO check server credentials are complete for nicer error messages
@@ -348,7 +348,7 @@ def _report_dry_run(
                 type="schema",
                 name="Check that blob files match the contract",
                 result=ResultEnum.warning,
-                reason="Dry run is incomplete: file-metadata checks for blob schemas are not planned.",
+                reason="Checks for Azure Blob storage are not considered in dry runs.",
                 engine="datacontract-cli",
             )
         )
--- a/datacontract/engines/data_contract_test.py
+++ b/datacontract/engines/data_contract_test.py
@@ -20,7 +20,7 @@ from datacontract.engines.datacontract.check_that_datacontract_contains_valid_se
 from datacontract.engines.fastjsonschema.check_jsonschema import check_jsonschema
 from datacontract.engines.ibis.ibis_check_execute import build_check_stubs, execute_ibis_checks, set_result
 from datacontract.model.exceptions import DataContractException
-from datacontract.model.run import ResultEnum, Run
+from datacontract.model.run import Check, ResultEnum, Run
 from datacontract.model.server import resolve_server_overrides
 
 
@@ -39,6 +39,7 @@ def execute_data_contract_test(
     filter: str | None = None,
     filters: dict[str, str] | None = None,
     metadata_only: bool = False,
+    dry_run: bool = False,
     config: Config | None = None,
     untrusted_contract: bool = False,
 ):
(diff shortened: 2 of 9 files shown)


OGsiji:
Pushed two fixes after testing the paths I had not exercised.

**An incomplete plan was invisible.** The blob-schema warning was being added correctly, but a dry run resolved to `skipped` unconditionally, so the summary read `⚪ no checks were executed` while the plan was quietly missing checks — the opposite of the requirement to make incompleteness obvious. A dry run now reports the most severe result its own checks carry:

```
🟠 data contract has warnings. Found the following warnings:
1) Check that blob files match the contract: Dry run is incomplete:
   file-metadata checks for blob schemas are not planned.
```

Still exits `0`. Two tests cover it.

**The JSON Schema condition was duplicated.** Deciding whether those checks apply was written out twice, once in the execution path and once in the dry run. That is the one place a plan could silently drift from what actually runs, which would defeat the point of the feature. Both now call the same predicate, so the execution path lost a copy rather than gaining one.

Also verified along the way, none of which needed changes: the plan matches a real run under `--checks` and `--dimension` filters (now asserted), Kafka and no-server contracts behave as they do without the flag, and `--output-format junit` reports the planned checks as skipped. The credentials test now clears the environment variables it assumes are absent, so it cannot block on a connection where a developer happens to have them set.

On the `3.13` job: it was cancelled at the 20 minute limit rather than failing an assertion, with orphan `pytest`/`uv` processes. I ran the whole suite on 3.13 locally — 2104 passed, and the only two failures were in `test_download_datacontract_file.py`, which pass on their own. Wall clock was 58 minutes at 5% CPU, so the time went on network stalls in those download tests, not on work. Nothing in this change makes a network call; it removes one, since a dry run no longer fetches the API response. Happy to be told otherwise if you have seen that job wedge differently.

OGsiji:
@jochenchrist

datacontract/engines/data_contract_test.py
```diff
@@ -105,6 +117,21 @@ def execute_data_contract_test(
...
+            server,
+            specs,
+            schema_name=schema_name,
+            check_categories=check_categories,
+            dimensions=dimensions,
+            quality_ids=quality_ids,
+            tags=tags,
+            config=config,
+        )
+        return
```
jschoedl, line 142: This needs to come after `if metadata_only` - the dry run should have the same filters as a real run

docs/docs/testing/index.md
```diff
@@ -123,6 +123,40 @@ Use `--metadata-only` to skip these value-level checks: only the schema-reading
...
+```
+
+```json
+{ "name": "Check that field order_total has no missing values",
+  "implementation": "missing_count(order_total) = 0",
+  "result": "skipped" }
+```
+
+Because nothing connects, a dry run needs no credentials for the server, which
+makes it usable on a pull request build that has no warehouse access. A dry run
```
jschoedl: That's not entirely true: A contract can reference semantics from an external source, in which case those are fetched and need to be accessible. (Docs: https://docs.datacontract.com/semantics)

datacontract/engines/data_contract_test.py
```diff
@@ -271,6 +290,71 @@ def get_server(data_contract: OpenDataContractStandard, server_name: str = None)
...
+    # The blob checks read file metadata to decide which checks exist at all,
+    # so they cannot be planned without reading. Say so rather than reporting a
+    # plan that is quietly missing them.
+    if server.type == "azure" and _has_blob_schemas(data_contract, schema_name):
+        run.checks.append(
+            Check(
+                type="schema",
+                name="Check that blob files match the contract",
+                result=ResultEnum.warning,
+                reason="Dry run is incomplete: file-metadata checks for blob schemas are not planned.",
```
jschoedl:
```suggestion
                reason="Checks for Azure Blob storage are not considered in dry runs.",
```

datacontract/command_test.py
```diff
@@ -217,6 +217,13 @@ def test(
             "Checks that read row values are skipped."
         ),
     ] = False,
+    dry_run: Annotated[
+        bool,
+        typer.Option(
+            help="Report the checks that would run, without connecting to the server or reading any data. "
+            "Every reported check has the result 'skipped'."
+        ),
```
jschoedl: this is not true, `warning` is possible as well

jschoedl (changes requested):
Thank you for the PR!

OGsiji:
Thanks — all four applied.

**The filter ordering was the real one.** You were right that it has to come after `--metadata-only`, and the effect was bigger than I expected: on the `local-json` fixture the plan listed 40 checks where a real `--metadata-only` run executes 19. A plan narrowed by only some of the filters is exactly the failure this feature is supposed to prevent, and I had built the guard for it on the JSON Schema side and then walked past it here.

Two tests now hold the plan and the run to the same filters, and the plan keeps each reason distinct, so `--metadata-only` skips read as their own thing rather than as dry-run skips.

**The help text** contradicted the warning I had added myself — fixed to mention `warning`.

**The semantics point** I had not considered at all. The docs now say a dry run is offline for the server rather than offline in general, and link to the semantics page, since the checks cannot be worked out until those are resolved.

**Wording suggestion** taken as written.

Also rebased onto main; the CHANGELOG entry moved to Unreleased now that 1.1.2 has shipped. Full suite green at 2198.

On #1531 — thanks for adding the dimensions, that was a real miss. I checked the new checks worked but not that they survived the filters that already existed, which is the same mistake as the ordering here. Noted.

jschoedl (approved):

PR #1528, opened 2026-08-22
fix: take the sql import server's database and schema from the table name

### LINKED ISSUE #651: Database and schema in Snowflake CREATE TABLE statement not picked up for servers section of YAML on import
Snowflake  SQL script has a `CREATE TABLE dbs.schema.table`, but the generated YAML reports errors on missing account, database, and schema. Account I understand, but database and schema are present and should be included in the `servers` section.

ReguiguiMohamed:
`datacontract import sql` parsed the catalog and schema of a qualified `CREATE TABLE` and then dropped them, so the generated `servers` block always carried the `my_database` and `my_schema` placeholders. They now come from the created tables.

The placeholders stay where the DDL identifies no single server: no qualifier, two created tables in different places, or a name carrying a substituted `${var}` token. Only `CREATE TABLE` counts, so a query source in a CTAS or a `CREATE VIEW` target cannot supply the location.

This applies to snowflake, sqlserver, postgres and redshift, where a qualified name means database.schema. On mysql, oracle, bigquery and databricks the parts mean something else and belong in other server fields, so their DDL is left to the placeholders.

Closes #651

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,6 +19,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - **breaking:** the `engine` field of test results is now always one of `datacontract-cli`, `dbt` or `jsonschema`; the values `datacontract`, `ibis` and `dbt-sync` no longer occur (#1505)
 
 ### Fixed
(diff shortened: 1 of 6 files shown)


jschoedl:
Hi, thank you for the PR! I added two minor changes, otherwise LGTM.

PR #1530, opened 2026-08-22
fix: import only the tables that CREATE TABLE statements create

### LINKED ISSUE #1529: `datacontract import sql` fails on a DDL file that contains `CREATE SCHEMA`
A DDL file that creates its schema before its tables cannot be imported.

`schema.sql`:

```sql
CREATE SCHEMA sales;
CREATE TABLE sales.orders (id INT);
```

```bash
datacontract import sql --source schema.sql --dialect postgres
```

Actual:

```
Error: 'NoneType' object has no attribute 'name'
```

Nothing is imported. Raised at `datacontract/imports/sql_importer.py:75`, `table_name = table.this.name`. `CREATE SCHEMA sales;` alone is enough, and `--dialect snowflake` fails the same way.

Expected: the `CREATE SCHEMA` is skipped and `sales.orders` imports, as it does when the file has no `CREATE SCHEMA` line.

Version 1.1.1, and `main` at f8059d9.

ReguiguiMohamed:
`datacontract import sql` failed with `AttributeError: 'NoneType' object has no attribute 'name'` on any DDL file containing `CREATE SCHEMA`.

The importer collected every table node beneath a `CREATE` rather than the table each `CREATE TABLE` creates, and `CREATE SCHEMA sales` parses to a table node with no table name. It now reads the target of each `CREATE TABLE`, which also drops the empty schema objects `CREATE DATABASE` and `CREATE VIEW` produced, and the duplicates from their query sources.

Fixes #1529

The four `test_import_sql_*` files pass. The full suite matches `main`, with one extra pass for the new test; the remaining failures are optional extras and Docker missing locally.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,6 +19,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - **breaking:** the `engine` field of test results is now always one of `datacontract-cli`, `dbt` or `jsonschema`; the values `datacontract`, `ibis` and `dbt-sync` no longer occur (#1505)
 
 ### Fixed
+- `datacontract import sql` no longer fails on a DDL file that contains `CREATE SCHEMA` (#1529)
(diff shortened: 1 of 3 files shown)


jschoedl:
Thanks!
I used the opportunity to fix a bug close to the code you edited.

Issue #1514, opened 2026-08-16
Some `logicalTypeOptions` are not enfoced


OGsiji:
While answering a question on the ODCS spec repo about declaring more than one date format ([bitol-io/open-data-contract-standard#306](https://github.com/bitol-io/open-data-contract-standard/issues/306)), I traced how `logicalTypeOptions.format` flows through the CLI and ended up somewhere I didn't expect. Filing it as a question rather than a bug, because the answer may well be "intentional".

### A declared date format is never applied

For a property with `logicalType: date` / `timestamp` / `time`, `logicalTypeOptions.format` holds a JDK `DateTimeFormatter` pattern (`yyyy-MM-dd`, per the ODCS schema's own examples). As far as I can tell it is never read:

- `engines/checks/create_checks.py` builds checks from `minLength`, `maxLength`, `minimum`, `maximum`, `exclusiveMinimum`, `exclusiveMaximum`, `pattern` and `enum`. `format` is not among them.
- `export/jsonschema_exporter.py` *does* read it, but `convert_type_format` substitutes a fixed value for these types and drops the declared one:

```python
if type_str.lower() in ["timestamp", "timestamp_tz", "date-time", "datetime"]:
    return "string", "date-time"
if type_str.lower() in ["date"]:
    return "string", "date"
if type_str.lower() in ["time"]:
    return "string", "time"
```

So `format: "dd/MM/yyyy"` on a date column has no effect on any server type.

That may be the right call — a JDK pattern has no JSON Schema equivalent, and `date-time` is the closest thing available. If so, it's worth saying in the docs, because the option looks enforceable and isn't.

### It behaves differently on a string property

For `logicalType: string`, `convert_type_format` passes the declared format straight through to JSON Schema, where `fastjsonschema` validates it. But `check_jsonschema` returns early unless `server.format == "json"`, so:

| property | `server.format: json` | any other server |
|---|---|---|
| `string` with `format` | validated | ignored |
| `date`/`timestamp`/`time` with `format` | ignored (overridden) | ignored |

Same contract option, three different fates depending on the logical type and the server. That's the part that seems most worth resolving one way or the other.

### Other options with no reader

While looking, these appear in the ODCS 3.1.0 `logicalTypeOptions` surface but have no handling in `export/` or `engines/`:

- `multipleOf` (integer, number) — written by the Excel importer, read by nothing
- `maxItems`, `minItems`, `uniqueItems` (array)
[...]

PR #1531, opened 2026-08-24
Check the ODCS array options minItems, maxItems and uniqueItems

OGsiji:
Part of #1514, where you said the unimplemented `logicalTypeOptions` would be welcome.

```yaml
- name: tags
  logicalType: array
  logicalTypeOptions:
    minItems: 1
    maxItems: 10
    uniqueItems: true
```

These measure the elements of one row's array rather than counting rows, so they are `invalid_count` checks like the existing options, and compile to `ARRAY_LENGTH` and `LIST_DISTINCT` through ibis. One check per constraint, so a failure names which one broke.

`uniqueItems: false` is the default and asserts nothing, so it produces no check rather than one that always passes.

### The case I want to flag

A column the contract calls an array, but the server does not.

My first version guarded the predicate on the column being an array and left the constraint off otherwise. That turned out to be wrong in a way worth describing: when `_invalid_expr` returns `None` the executor records `invalid_count = 0 (no validity constraints configured)` and **evaluates the check as passed**. So a contract asserting `minItems: 2` against a `VARCHAR` column would have reported green.

It now reports an error naming the type it found:

```
Column tags is string, not an array, so the constraint cannot be measured.
```

A constraint that cannot be measured is not a constraint that holds. Worth knowing that `None` path exists, since anything else added to `_valid_expr` that can decline to build a predicate inherits the same behaviour.

Contract-level mistakes are already caught upstream — the ODCS schema rejects `minItems` on a string property, so this only arises when the column itself has drifted.

### On `minProperties` / `maxProperties`

I said on the issue I would do these too, and I have left them out deliberately.

They are only row-level on a `map` column, where the key count varies per row and ibis gives a usable `.length()`. On a `struct` the fields are fixed by the schema, so the count is the same for every row and the check is either trivially true or trivially false for the whole table — closer to the existing physical type check than to a value check.

So they are two different checks wearing one option name, and which you get depends on the physical type rather than anything the contract says. I would rather ask than guess: should `minProperties` apply only to map-typed columns, become a schema-level assertion on structs, or be left alone? Happy to implement whichever you prefer.

`multipleOf` and `defaultTimezone` are also still unimplemented; I skipped them here to keep this reviewable.

12 tests added, full suite green (2120 passed).


jschoedl:
Thanks! I applied 3 small changes immediately, hope that's okay :)

OGsiji:
Hey @jschoedl, thanks for the additions, Saw this late.

Issue #1358, opened 2026-07-03
Setting DATACONTRACT_BIGQUERY_BILLING_PROJECT causes test failures


kien-truong:
When `DATACONTRACT_BIGQUERY_BILLING_PROJECT` is set, the `project` configuration in servers are ignored and all tests use the value of `DATACONTRACT_BIGQUERY_BILLING_PROJECT` as the BigQuery datasets project, causing test failures.

kien-truong:
This is because for ibis, you probably need to set the `catalog` value to the BigQuery server's `project`, otherwise it'd use the default credential project_id.

https://ibis-project.org/backends/bigquery#ibis.backends.bigquery.Backend.list_databases

jochenchrist:
related: #1291

jochenchrist:
What would be your expected behavior?

kien-truong:
The correct behavior is as described in ticket #1291 
However, pr #1292 that closed it doesn't work correctly, and it also doesn't have any validation tests.

simonharrer:
Can you try 1.0.16? I think I have fixed it now.

jschoedl:
I could still reproduce this on 1.0.16. Will take a look next week.

PR #1540, opened 2026-08-25
fix: qualify BigQuery dataset_id with the data project when billing separately

aabf:
## Summary
- `DATACONTRACT_BIGQUERY_BILLING_PROJECT` lets a run bill/execute BigQuery jobs under a different project than the one holding the data. The bigquery branch in `connect.py` builds a `bigquery.Client(project=billing_project)` and passes it into `ibis.bigquery.connect(project_id=project, dataset_id=dataset, client=client)` — with `dataset_id` unqualified.
- `ibis.backends.bigquery.Backend.do_connect` prefers `client.project` over the `project_id` argument whenever a `client` is passed in (`client_project_id or project_id or default_project_id`). Since the client's project is the billing project, `parse_project_and_dataset` ends up splitting `billing_project` as *both* the data project and the billing project — the `project_id` we asked for is silently discarded.
- Net effect: every query looks for the dataset in the billing project instead of the data project, e.g. `Not found: Dataset <billing_project>:<dataset>`.
- Fix: qualify `dataset_id` as `f"{project}.{dataset}"` in that branch, so `parse_project_and_dataset` splits data project vs. billing project correctly regardless of which project the client bills to.

## Test plan
- [x] Added `test_billing_project_client_uses_the_impersonated_credentials` assertion covering the qualified `dataset_id`
- [x] `pytest tests/test_connect_bigquery.py` passes (5/5)

### DIFF
--- a/datacontract/engines/ibis/connections/connect.py
+++ b/datacontract/engines/ibis/connections/connect.py
@@ -162,9 +162,12 @@ def connect_ibis(
             from google.cloud import bigquery as bq_client_lib
 
             client = bq_client_lib.Client(project=billing_project, credentials=credentials)
+            # ibis's do_connect prefers client.project over project_id when a
+            # client is passed, so qualify dataset_id or it resolves against
+            # billing_project instead of project.
             return ibis.bigquery.connect(
                 project_id=project,
-                dataset_id=dataset,
+                dataset_id=f"{project}.{dataset}",
                 client=client,
             )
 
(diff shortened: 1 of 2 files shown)


aabf:
Closing as duplicate -- verified this is already fixed on `main` in a6ebfab4 (#1524, fixing #1358), merged 2026-08-21: same root cause and approach (drop the pre-built `client`, qualify `dataset_id` as `<project>.<dataset>`), plus more (validation errors for missing project/dataset, credentials reaching the Storage Read API client too). Ran `test_billing_project_keeps_the_contract_project_as_the_data_project` against the real ibis backend to confirm. It's under `## [Unreleased]` in CHANGELOG.md, just not released yet. Sorry for the noise!

PR #1542, opened 2026-08-26
feat: Add status 'draft' default value when converting DCS to ODCS

michal-swiatowy:
Fixes https://github.com/datacontract/datacontract-cli/issues/1532

- [x] Tests pass (`uv run pytest`)
- [x] Code formatted (`uv run ruff check --fix && uv run ruff format`)
- [x] Docs updated (if relevant)
- [x] CHANGELOG.md entry added

PR #1543, opened 2026-08-26
fix(export): populate required ODCS status field when source contract omits it

### LINKED ISSUE #1532: `status` property missing when exporting to ODCS
Hi there. I'm running a migration project from DCS 1.2.0 spec to ODCS 3.1.0. I'm using datacontract-cli 1.0.16. 

When running the following: `datacontract export odcs <dcs> --output <odcs>` and running linting on `<odcs>` I get:
```
╭────────┬────────────────────────────────────────┬───────┬─────────────────────────────────────────╮
│ Result │ Check                                  │ Field │ Details                                 │
├────────┼────────────────────────────────────────┼───────┼─────────────────────────────────────────┤
│ failed │ Check that data contract YAML is valid │       │ data must contain ['status'] properties │
╰────────┴────────────────────────────────────────┴───────┴─────────────────────────────────────────╯
```

Is there any way to populate default `status` value or it needs to be done as part of post-process?

Thanks in advance.

britsync07-prog:
## Summary

Fixes #1532

Exporting a Data Contract Specification (DCS) file to ODCS produced YAML that failed ODCS 3.1.0 validation with:

`	ext
data must contain ['status'] properties
`

**Root cause:** convert_dcs_to_odcs() only set odcs.status when info.status was present in the source contract. Since DCS does not require status, many valid contracts exported without it - and status is a required top-level property in the ODCS schema.

## Changes

- datacontract/imports/dcs_importer.py: default odcs.status to "draft" (a value from the ODCS status vocabulary) when the source contract omits it. Explicit values are still passed through unchanged.
- 	ests/test_export_odcs_v3.py: add regression test covering both cases.

## Validation

- Contract without info.status now exports with status: draft
- Contract with explicit info.status: active still exports status: active
- Exported output now satisfies the required fields of the bundled odcs-3.1.0.schema.json

### DIFF
--- a/datacontract/export/odcs_v3_exporter.py
+++ b/datacontract/export/odcs_v3_exporter.py
@@ -14,13 +14,13 @@ def to_odcs_v3_yaml(data_contract: OpenDataContractStandard) -> str:
     """Export the internal ODCS model to YAML format.
 
     Since the internal model is now ODCS, this is a simple serialization.
+    ODCS requires a top-level `status`; default to 'draft' when omitted.
     """
-    return data_contract.to_yaml()
+    return to_odcs_v3(data_contract).to_yaml()
 
 
 def to_odcs_v3(data_contract: OpenDataContractStandard) -> OpenDataContractStandard:
-    """Return the internal ODCS model.
-
-    Since the internal model is now ODCS, this is an identity function.
-    """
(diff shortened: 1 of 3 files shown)


jschoedl:
Fixed in #1542.

PR #1547, opened 2026-08-28
Give each servicelevel freshness/retention check a unique key

### LINKED ISSUE #1515: No result found for the second slaproperty check
When running the `datacontract test` command, only the result of the first test is displayed (I noticed this during the freshness checks). From the second test onwards, the result is always NULL.

Is this a bug, or am I misunderstanding something?

Datacontract 

### Example slaProperties:

slaProperties:
  - id: monitoring_airflow_dag_task_contract_freshness
    property: freshness
    value: 4
    unit: hour
    element: monitoring_airflow_dag_task_contract.dag_run_start_date
    description: The DAG metadata history must be up to date within 4 hours.
  - id: monitoring_data_contract_check_freshness
    property: freshness
    value: 4
    unit: hour
    element: monitoring_data_contract_check.check_timestamp
    description: The most recent execution of a Data Contract Check must not be more than 4 hours old.

### Output datacontract cli:
╭────────┬───────────────────────────────────────────────────────────────────────────┬──────────────────────────────────┬─────────╮
│ Result │ Check                                                                     │ Field                            │ Details │
├────────┼───────────────────────────────────────────────────────────────────────────┼──────────────────────────────────┼─────────┤
│        │ Freshness of monitoring_data_contract_check.check_timestamp < 4h          │ monitoring_data_contract_check.… │         │
│ passed │ Check that model monitoring_airflow_dag_task_contract has row_count >= 1  │ monitoring_airflow_dag_task_con… │         │
│ passed │ Check that field 'contract_content_hk' is present                         │ monitoring_airflow_dag_task_con… │         │
│ passed │ Check that field contract_content_hk has physical type VARCHAR            │ monitoring_airflow_dag_task_con… │         │
│ passed │ Check that field contract_content_hk has no missing values                │ monitoring_airflow_dag_task_con… │         │
│ passed │ Check that field 'dag_hk' is present                                      │ monitoring_airflow_dag_task_con… │         │
│ passed │ Check that field dag_hk has physical type VARCHAR                         │ monitoring_airflow_dag_task_con… │         │
│ passed │ Check that field dag_hk has no missing values                             │ monitoring_airflow_dag_task_con… │         │
│ passed │ Check that field 'dag_id' is present                                      │ monitoring_airflow_dag_task_con… │         │
│ passed │ Freshness of monitoring_airflow_dag_task_contract.dag_run_start_date < 4h │ monitoring_airflow_dag_task_con… │         │
│ passed │ Check that field 'dag_run_state' is present                               │ monitoring_airflow_dag_task_con… │         │
│ passed │ Check that field dag_run_state has physical type VARCHAR                  │ monitoring_airflow_dag_task_con… │         │
...
╰────────┴───────────────────────────────────────────────────────────────────────────┴──────────────────────────────────┴─────────╯

erikgrip2:
Fixes #1515

With more than one freshness or retention promise in a contract, every check got the same key (`servicelevel_freshness` / `servicelevel_retention`). Results are written to the first check with a matching key, so all outcomes landed on the first check and the others stayed without a result.

This keys each check by its model and field, following the convention the schema checks already use (`{model}__{field}__{check_type}`), in both the engine check path and the SodaCL export path, which intentionally use the same key strings.

**Note to reviewers:** this changes the `key` field in the test results output.
- Before: every freshness check had `key: servicelevel_freshness` and every retention check `key: servicelevel_retention`, regardless of what it checked.
- After: each check's key names the table and column it checks, e.g. `key: orders__order_timestamp__servicelevel_freshness`.

**Not fixed here:**
- In the SodaCL export, two retention promises on the same table still produce two metrics with the same name (`{model}_servicelevel_retention`), which collide inside Soda. That looks like a separate, smaller fix.
- Two promises of the same kind on the same column (say, two freshness promises on `orders.order_timestamp`) would still share one key. This could be covered too, for example by numbering the checks the way the quality checks already do, but I'm not sure it's worth adding logic for a case no one has reported. Happy to include it if you'd like.

Verified with `pytest` (2188 passed), `ruff check .`, `ruff format --check .`, `git diff --check`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - `datacontract import sql` takes the server's `database` and `schema` from a qualified `CREATE TABLE`, instead of always writing placeholders (#651)
 - `datacontract import sql` no longer fails on a DDL file that contains `CREATE SCHEMA` (#1529)
 - `datacontract test` now supports ISO 8601 retention periods correctly (previously, only the first component was considered) (#1538)
+- `datacontract test` reports each freshness and retention check result on its own check, instead of writing every result to the first one (#1515)
 
 ## [1.1.2] - 2026-08-26
 
--- a/datacontract/engines/checks/create_checks.py
+++ b/datacontract/engines/checks/create_checks.py
@@ -909,7 +909,7 @@ def _freshness_check(
         return None
 
     return CheckSpec(
-        key="servicelevel_freshness",
+        key=f"{model}__{field}__servicelevel_freshness",
         category="servicelevel",
         type="servicelevel_freshness",
         name=f"Freshness of {model}.{field} < {sla.value}{unit[0]}",
@@ -934,7 +934,7 @@ def _retention_check(
     if seconds is None:
         return None
     return CheckSpec(
-        key="servicelevel_retention",
+        key=f"{model}__{field}__servicelevel_retention",
         category="servicelevel",
         type="servicelevel_retention",
         name=f"Retention of {model}.{field} < {seconds}s",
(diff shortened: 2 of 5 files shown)


jschoedl:
Thanks!

PR #1554, opened 2026-09-01
feat(kafka): add DATACONTRACT_KAFKA_GROUP_PREFIX to customise consumer group ID

### LINKED ISSUE #1553: [feature request, Kafka test] Add custom group.id prefix to Kafka env variables
Dear datacontract CLI team,

I'm facing an issue in a regulated internal Confluent platform offering that might occur in other environments/companies as well. When testing a Kafka datacontract using `datacontract test`, I keep getting a `group_authorization` error because my service account is not allowed to read from consumer groups that are prefixed with `datacontract-cli-*`. 

My service accounts are bound to specific prefixes for billing purposes, and I cannot change this without raising internal exception requests. 

Here's my proposal: 
Can we add a "custom group prefix" environment variable (e.g., `DATACONTRACT_KAFKA_GROUP_PREFIX` that is then added instead of or before the `datacontract-cli-<guid>`?
Or just add `DATACONTRACT_KAFKA_GROUP_ID` and always use that group if set?

In any case, the documentation should show which permissions/ACLs the service account must have so others won't run into this issue.

Thanks and best regards
Philipp

philipp-lutz:
## Summary

- Adds `DATACONTRACT_KAFKA_GROUP_PREFIX` env variable to allow users to customise the Kafka consumer group prefix used during `datacontract test`
- Default behaviour (`datacontract-cli-<uuid>`) is unchanged when the variable is unset
- A UUID is always appended to the supplied prefix, so each run still gets a unique group ID and offsets are never committed

## Motivation

On regulated Confluent platforms, service accounts are restricted to specific consumer group prefixes for billing/ACL compliance. The hard-coded `datacontract-cli-` prefix is unusable without a formal exception request. See #1553.

## Changes

- `datacontract/config/settings.py` — new `kafka_group_prefix` field + `get_kafka_group_prefix()` accessor
- `datacontract/engines/ibis/connections/kafka.py` — `_consumer_config()` uses the prefix from config
- `docs/docs/reference/kafka.md` — new row in the "Reading the topic" table
- `CHANGELOG.md` — entry under `[Unreleased]`
- `tests/test_test_kafka_decoding.py` — two unit tests for the default and custom prefix paths

## Test plan

- [x] `pytest tests/test_test_kafka_decoding.py` — all 37 tests pass (no broker needed)

Closes #1553


jschoedl:
Thanks, LGTM! (Apart from the CRLF line endings in the docs file that I fixed :)

PR #1555, opened 2026-09-02
fix(export): use physicalName as field name in jsonschema and avro exports (#1494)

### LINKED ISSUE #1494: physicalName being ignored when exporting to jsonschema and avro
Dear team, 

I've noticed that when exporting an ODCS schema to avro or jsonschema, it uses the `name` of the objects and properties and ignores the `physicalName`. 

Commands used:

- `datacontract export jsonschema --output jsonschema.json datacontract.yaml`
- `datacontract export avro --output avro.json datacontract.yaml`

ODCS Property:

```yaml
  - name: Equipment Name
    physicalName: name
    logicalType: string
    description: Equipment name based on the ISA-95 model
``` 

Result (avro):

```json
    {
      "name": "Equipment Name",
      "doc": "Equipment name based on the ISA-95 model",
      "type": "string"
    },
``` 

Expected result:

```json
    {
      "name": "name",
      "doc": "Equipment name based on the ISA-95 model",
      "type": "string"
    },
```

This problem extends to any `datacontract test` that checks a JSON using the json schema (e.g., when testing an HTTP API).

philipp-lutz:
## What

Fixes #1494.

`datacontract export jsonschema` and `datacontract export avro` emitted each field using its logical `name` and never consulted `physicalName`. A property with `name: "Equipment Name"` and `physicalName: "name"` was exported as `"Equipment Name"` instead of `"name"`. This also affected `datacontract test` flows that validate JSON against a generated JSON Schema.

## Change

Apply the established `prop.physicalName or prop.name` idiom (already used by the SQL, BigQuery, and SodaCL exporters) so the physical field name is used when set, falling back to the logical name otherwise:

- `datacontract/export/jsonschema_exporter.py` — `to_properties` (property key) and `to_required` (required entry); nested objects recurse through the same functions, so they're covered too.
- `datacontract/export/avro_exporter.py` — `to_avro_field` field name.

Left untouched (deliberately): Avro nested-record/enum names and JSON Schema `patternProperties` detection — those are type identifiers / regex heuristics, not field names.

## Tests

Added a focused test to `tests/test_export_jsonschema.py` and `tests/test_export_avro.py` asserting `physicalName` wins and `name` is the fallback when it is unset.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,7 +14,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - `datacontract test --dry-run` reports the checks a run would execute without connecting to the server or reading any data (#1510)
 
 ### Fixed
-- `datacontract export jsonschema` and `datacontract export avro` use a property's `physicalName` as the field name when set, instead of the logical `name` (#1494)
+- `datacontract export jsonschema`, `datacontract export avro` and `datacontract test` on local files use a property's `physicalName` as the field name when set, instead of the logical `name` (#1494)
 - `datacontract import sql` takes the server's `database` and `schema` from a qualified `CREATE TABLE`, instead of always writing placeholders (#651)
 - `datacontract import sql` no longer fails on a DDL file that contains `CREATE SCHEMA` (#1529)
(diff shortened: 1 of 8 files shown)


jschoedl:
Thanks, LGTM! I used the opportunity to fix a similar issue with `datacontract test` for local files.

PR #1605, opened 2026-09-08
Use named loggers instead of the root logger

### LINKED ISSUE #1604: Use named loggers instead of the root logger
datacontract-cli currently uses root logging functions such as:

```python
logging.info(message)
logging.warning(message)
logging.error(message)
```

This makes it difficult for applications embedding datacontract-cli to configure its logs independently from application logs.

Would you consider using module-level named loggers throughout the package?

```python
logger = logging.getLogger(__name__)
``` 

For example:

```python
logger.info(message)
logger.warning(message)
logger.error(message)
```

This would create loggers under the shared `datacontract` namespace, such as `datacontract.model.run`, allowing consumers to configure the whole library:

```python
logging.getLogger("datacontract").setLevel(logging.CRITICAL)
```
or configure individual modules separately.

Keeping the default `propagate=True` should preserve the current behavior for users relying on root handlers while providing applications with finer logging control.

I’d be happy to submit a pull request replacing the root logging calls with module-level named loggers if this approach is acceptable

sondrfos:
Replace root logger calls with named module loggers throughout the CLI.

This makes log output easier to trace and lets consumers configure logging behavior per module.

Closes #1604


datacontract/engines/ibis/connections/kafka.py
```diff
@@ -54,6 +54,8 @@
 
 _SASL_MECHANISMS = ("PLAIN", "SCRAM-SHA-256", "SCRAM-SHA-512")
 
+logger = logging.getLogger(__name__)
+
```
jschoedl:
```suggestion
```
Already declared in L39

jschoedl (changes requested):
Thanks, looks good! There are 4 root-logger calls left (search for `logging.warning`):
- 2 in `datacontract/api.py`
- 2 in `datacontract/lint/resolve.py`

PR #1643, opened 2026-09-21
fix: preserve nested Spark varchar types during import (#1634)

### LINKED ISSUE #1634: Spark import loses varchar length inside struct, producing a contract that fails Databricks validation
## Summary

A Databricks table column with a native type like:

`struct<varchar_field:varchar(100)>`

is imported by the Spark importer as:

```yaml
physicalType: struct<varchar_field:string>
```

This produces a contract that fails validation against the same table, even though the original source has a length-constrained varchar inside a struct.

This is a regression relative to the workaround that was previously implemented for affected schema type checks.

## Reproduction

Minimal table example:

```sql
CREATE TABLE test_struct_varchar (
  id INT,
  payload STRUCT<varchar_field:VARCHAR(100)>
);
```

Generated YAML after import:

```yaml
version: 1.0
id: test-struct-varchar
type: table
description: ""
schema:
  fields:
    - name: id
      type: integer
      physicalType: int
    - name: payload
      type: struct
      physicalType: struct<varchar_field:string>
      fields:
        - name: varchar_field
          type: string
          physicalType: string
```

Validation against the source table fails with:

```text
expected physical type 'struct<varchar_field:string>' but the column is 'struct<varchar_field:varchar(100)>'
```

Manually corrected YAML:

```yaml
version: 1.0
id: test-struct-varchar
type: table
description: ""
schema:
  fields:
    - name: id
      type: integer
      physicalType: int
    - name: payload
      type: struct
      physicalType: struct<varchar_field:varchar(100)>
      fields:
        - name: varchar_field
          type: string
          physicalType: varchar(100)
```

This corrected version validates successfully.

## Root cause

The Spark dataframe schema normalizes nested `varchar(n)` values inside complex types to Spark `string`, losing the length information before the upstream importer records the contract.

The complete native type is still available in the Databricks metadata returned by `DESCRIBE`, but the Spark importer uses the lossy schema representation instead of the authoritative native type.

This library already contains a local mitigation for top-level varchar columns in `set_varchar()`, but it does not handle complex nested types like `struct`, `array`, or `map` when they contain `varchar(n)`.

## Versions

- datacontract-cli: 1.2.0
- Python: [fill in]
- PySpark: [fill in]
- Databricks Runtime: [fill in]
- Databricks Connect: [fill in]

## Relevant upstream history

- #1048: Spark imports emit native SQL names using `dataType.simpleString()`
- #1219: reported maps / nested types
- #1245: reported `varchar(n)` nested in Databricks structs
- #1253: shipped in 0.12.5 and skipped affected schema type checks as a workaround

## Expected behavior

The imported contract should remain valid against the same source table, preserving `varchar(n)` when the native Databricks metadata still exposes the precise type.

A contract produced by the Spark/Databricks import should not fail validation solely because nested varchar lengths were normalized away in the Spark schema.

## Notes
This issue is best treated as a regression or incomplete replacement of the 0.12.5 workaround.

IchEssBlumen:
fixes #1634


jschoedl:
Thank you! Have you considered reading the type from the field metadata instead? Spark keeps the original type string under `__CHAR_VARCHAR_TYPE_STRING` when it widens char/varchar:

```
payload struct<varchar_field:string,n:int> {'__CHAR_VARCHAR_TYPE_STRING': 'struct<varchar_field:varchar(100),n:int>'}
```

`_property_from_struct_type` could take physicalType from there and split it alongside the DataType.

IchEssBlumen:
@jschoedl Good point. I have refactored the code and now it got much simpler. Please review again.

datacontract/imports/spark_importer.py
```diff
@@ -1,10 +1,22 @@
...
+try:
+    # pyspark is deliberately not a package dependency (see pyproject.toml) since a
+    # Spark import only ever runs against a session the caller already built, so
+    # pyspark is necessarily importable in that process. This module is still
+    # importable without it (e.g. for unit-testing the pure-Python helpers below);
+    # `DataFrame`/`SparkSession`/`types` are only dereferenced once real Spark
+    # objects are involved.
+    from pyspark.sql import DataFrame, SparkSession, types
+except ImportError:
+    DataFrame = SparkSession = types = None
```
jschoedl, line 19: I don't think we need this. The importer is only loaded lazily when someone runs `import spark`, and there's no test that imports the helpers without pyspark. It also makes the error worse when pyspark is missing: `datacontract import spark` used to fail with `Module spark could not be loaded.`, and now it fails with `'NoneType' object has no attribute 'builder'`.

jschoedl (changes requested):
Thanks, looks good!

I pushed a few commits on top:
- Array elements and map keys/values now keep their `varchar`/`char` types as well, e.g. `array<varchar(10)>` and `map<string,struct<z:varchar(13)>>`. Before, only struct children did.
- Top-level `map` columns now import with their real key and value types. They were always `string`/`string` before, even for `map<string,bigint>`.

datacontract/imports/spark_importer.py
IchEssBlumen: I have removed the try/except block again. And I tested everything including your changes.  Looks all good.

PR #1649, opened 2026-09-22
Add OData import support

jahlen:
Adds OData 4.x import from CSDL XML or JSON, using URLs or local files. Supports selecting individual EntitySets or importing all advertised EntitySets, with authentication through the existing configuration system.

## Sample usage
```bash
datacontract import odata \
  --service-root-url 'https://xmart-api-public-uat.who.int/refmart/' \
  --entity-set ref_country \
  --output datacontract.yaml
```

## Sample generated datacontract
```yaml
servers:
- server: source
  type: api
  customProperties:
  - property: apiType
    value: odata
  - property: odataVersion
    value: '4.0'
  - property: odataMetadataUrl
    value: https://xmart-api-public-uat.who.int/refmart/$metadata
  - property: format
    value: json
  location: https://xmart-api-public-uat.who.int/refmart/
schema:
- name: REF_COUNTRY
  physicalType: object
  customProperties:
  - property: odataEntitySet
    value: REF_COUNTRY
  - property: odataEntitySetUrl
    value: https://xmart-api-public-uat.who.int/refmart/REF_COUNTRY
  logicalType: object
  physicalName: REF_COUNTRY
  properties:
  - name: GEO_M49_CODE
    physicalType: Edm.String
    logicalType: string
    required: false
  - name: CODE_ISO_2
    physicalType: Edm.String
    logicalType: string
    required: false
    # More properties...
```

I would suggest to add "odata" as a separate server type in the ODCS so that it can be handled without custom properties. Meanwhile, I have used the server type "api".

## Current limitations
- This feature implements **import only**.
- Complex types, collection-valued fields, entity or container inheritance, enum types, type definitions and
  types outside the table produce errors when used by selected schemas.


jschoedl:
Hi, thank you for the PR!

The docs and tests are pretty long, can you trim them down to what is realistically needed? You can take a look at how the other importers are documented and tested for reference. (In particular, descriptions of the development process like "`odataEntitySet` now belongs to each schema object" should not be part of the docs - what does "now" refer to?)

jahlen:
Hi Jakob,

I've updated the docs and tests now. Let me know what you think!

datacontract/imports/odata_importer.py
```diff
@@ -0,0 +1,513 @@
...
+            CustomProperty(property="apiType", value="odata"),
+            CustomProperty(property="odataVersion", value=version),
+            CustomProperty(property="odataMetadataFile", value=str(metadata_file))
+            if metadata_file is not None
+            else CustomProperty(property="odataMetadataUrl", value=metadata_url),
+            # ODCS API servers do not allow a top-level format field.
+            CustomProperty(property="format", value="json"),
+        ]
+        contract.servers = [server]
+        return contract
```
jschoedl, line 130: Please call `report_unmapped_types(contract)` before returning (imported from `datacontract.imports.odcs_helper`): Some types are not defined in `ODATA_4_TYPES` (namely `Edm.Binary`, `Edm.Duration`, `Edm.Stream`, and the 16 `Edm.Geography*`/´Edm.Geometry` types), and such types should warn rather than fail.

(If you want, you can of course add them to the map in addition, but that's no hard requirement for a first version)

datacontract/imports/odata_importer.py
```diff
@@ -0,0 +1,513 @@
...
+    version = header.strip() if header is not None else document_version
+    if any(
+        not isinstance(value, str) or re.fullmatch(r"4\.[0-9]+", value) is None for value in (document_version, version)
+    ):
+        raise _schema_error(
+            f"Unsupported OData version (header={header!r}, document={document_version!r}); "
+            "expected an OData 4.x version in the form 4.[0-9]+."
+        )
+    if version != document_version:
+        raise _schema_error(f"Conflicting OData versions: header={version}, document={document_version}.")
```
jschoedl, line 258: Do we need such a check, or can't we just take `odataVersion` from the document and ignore the header?


I found that e.g. Northwind V4 sends `OData-Version: 4.0;`, which fails the `4.[0-9]+` check. Or TripPin RESTier answers with `OData-Version: 4.01` for 4.0 CSDL when the request allows 4.01, which would fail the conflict check.

datacontract/command_import.py
```diff
@@ -909,3 +909,54 @@ def import_athena(
...
+        ),
+    ] = None,
+    output: output_option = None,
+    owner: owner_option = None,
+    id: id_option = None,
+    debug: debug_option = None,
+):
+    """Import a data contract from OData 4.x CSDL XML or JSON metadata, using a URL or local file."""
+    if metadata_url is not None and metadata_file is not None:
+        raise typer.BadParameter("--metadata-url and --metadata-file are mutually exclusive.")
```
jschoedl, line 951: This check is duplicated in the importer (`odata_importer.py:59`), which also covers the Python API. Could be dropped here

datacontract/command_import.py
```diff
@@ -909,3 +909,54 @@ def import_athena(
...
     _write_result(result, output)
+
+
+@import_app.command(
+    name="odata",
+    epilog="Example: datacontract import odata --service-root-url https://example.com/odata/ --entity-set Products --output datacontract.yaml",
+)
+def import_odata(
+    service_root_url: Annotated[str, typer.Option(help="HTTP(S) root URL of the OData service.")],
+    service_root_file: Annotated[
```
jschoedl, line 920: Maybe name `service_document_file` to avoid confusion with `service_root_url`?

datacontract/imports/odata_importer.py
```diff
@@ -0,0 +1,513 @@
...
+        if valid:
+            parsed.port  # Validate a supplied port as well as the hostname.
+        if root and valid and ("?" in url or "#" in url):
+            valid = False
+    except ValueError:
+        valid = False
+    if not valid:
+        raise _schema_error(
+            f"Invalid {label}: expected an HTTP(S) URL without embedded credentials. "
+            "The service root must not contain a query or fragment."
```
jschoedl, line 153: This message is also used for --metadata-url and EntitySet URLs, where the sentence about the service root doesn't apply

jschoedl (changes requested):
Nice!

jochenchrist (approved):
LGTM (appart from jakobs remarks)

datacontract/imports/odata_importer.py
jahlen: Fixed!

datacontract/imports/odata_importer.py
jahlen: I've removed the check and now rely fully on the Version from the CSDL.

datacontract/command_import.py
jahlen: Removed the check.

datacontract/command_import.py
jahlen: Changed the name.

datacontract/imports/odata_importer.py
jahlen: Changed the error message.

